PR #16, opened 2018-11-25
Filled out eLevelId enum

nicholasdechiara:
- Added ELevelIdHelper class which contains code that generates the eLevelId enum from the mpq data
- Renamed some enum values for consistency (e.g. Act1_Town -> Act1_Town1, as it is in the mpq)

Because this enum is hard coded, this function generates the code that should be placed within; it's not a runtime generation. Depending on what we want to support in the future / what this enum is used for (e.g. if we wanted to support mods with custom level ids), we may want to consider using a different structure such as a dictionary that can be generated at runtime. For the moment, however, this seems fine.

Also, I renamed some of the enum values (Act1_Town -> Act1_Town1 and Act5_Baal_Entrance -> Act5_BaalEntrance) to be consistent with the auto-generation.

- Added OpenDiablo2.Core.UT unit test project
- Added a unit test that verifies EngineDataManager works

Format I'm used to for unit testing is to give each project a corresponding .UT project. If you'd prefer something else, I don't have strong feelings about it. I added a unit test for the EngineDataManager that tests loading the MPQ files. Currently, though, it requires you to put them in a specific place or to modify a parameter in the unit test code in order to locate your MPQ files. I've been trying to think of a nicer solution to this but haven't come up with anything yet (if we had code that tried to auto-locate the mpqs on the user's disk, that might be an option).

- Added a unit test that runs the ELevelIdHelper generate function

It's probably kind of messy to have the hook for this helper function be in a unit test, but it also doesn't make sense to call it anywhere in runtime as the enum can't be dynamically generated. Currently what I'm doing is debugging into the unit test and copy/pasting the result into the eLevelId code. Seeing as the default mpqs aren't going to change on us, this probably never needs to be run again.

### DIFF
--- a/OpenDiablo2.Common/Enums/eLevelId.cs
+++ b/OpenDiablo2.Common/Enums/eLevelId.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using OpenDiablo2.Common.Models;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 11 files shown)


nicholasdechiara:
Added a TestConsole which can run the elevelids enum generation and output the result to a file, and removed the unit test that was doing this (since it wasn't really a test)

PR #48, opened 2018-12-12
Load Enemy Type Info from monstats.txt

nicholasdechiara:
**Overview**

- _monstats.txt_ holds a huge (240 columns) table of info that defines many characteristics about monsters, npcs, hirelings, and other mobs. 
- Each row constitutes a different enemy definition
- EngineDataManager has been modified to load and store this information as a list of EnemyTypeConfigs
- EnemyState's constructor has been modified so that it requires an EnemyTypeConfig, which is used to define most of its stats.

**Loading the Data**

- ResourcePaths has a new path: MonStats --> _data\global\excel\monstats.txt_
- EngineDataManager has a new method, LoadEnemyData, which is called in initialization and should be where all load requests related to enemies are put.
- This method calls LoadEnemyTypeConfig(), which populates the EDM's EnemyTypeConfigs list with EnemyTypeConfig objects made from the _monstats.txt_
- Not all info from the table is loaded, I have tried to prune out info that exists only for reference or that is otherwise unused. **As we work we may discover that other info actually is needed, or (more likely) that some info isn't actually used.**

**Structure of the Data**

- EnemyTypeConfig stores basic info related to the enemy, such as its name, type, size, etc.
- It also stores the following objects: EnemyTypeAppearanceConfig, EnemyTypeCombatConfig, and three EnemyTypeDifficultyConfigs (one for each difficulty, normal, nightmare, and hell)
- EnemyTypeAppearanceConfig stores info related to the rendering of the enemy, such as whether or not it has certain animations, or if it has a light surrounding it.
- EnemyTypeCombatConfig stores info related to how the enemy acts in combat, such as what missiles it uses, whether it does elemental damage, what skills it has, etc.
- EnemyTypeDifficultyConfig stores info that is difficulty-specific, such as its level, hp, and resistances.

**Using the Data**

- Currently the data is not used anywhere during runtime. 
- The EnemyState's constructor calls for an EnemyTypeConfig, which would need to be pulled out of the EDM's store based upon some criteria (name of the mob?). 
- It also requires an eDifficulty flag, which indicates what difficulty the game is on. The EnemyTypeConfig object has a method GetDifficultyConfig(eDifficulty Difficulty) which gives you the appropriate difficulty config for a given game difficulty.

**Future Work**

- We will need to store the game's difficulty level somewhere (in the server?). When an enemy is created, it needs to be handed off this difficulty setting.
- EnemyState should be expanded to utilize more aspects of the EnemyTypeConfig, such as handling attacks, skills, item drops, etc.
- EnemyTypeConfig requires additional testing in order to understand what some of its data actually affects in the game.
- We should consider useful ways to index and categorize this data, as currently it is all thrown in a list. There doesn't seem to be a unique id for enemy definitions in the table itself.
- We should look into understanding how the AIParams work


essial:
Please resolve the merge conflicts so we can merge this in. We were in the middle of a huge equipment overhaul when this was posted.

PR #52, opened 2018-12-15
Address Patch MPQ Issue

nicholasdechiara:
**Overview**

- Diablo 2 uses a Patch_d2.mpq file to make changes to other existing files
- We need to load this mpq and allow it to overwrite existing data files in order to use the most recent versions of the data
- I have modified the mpqProvider in order to make it load the patch mpq last and use its files preferentially.
- The CharStats and Level Details needed to be changed as their patched files have different columns than the pre-patch files did. 

**FileList**

- An issue with the patch mpq is that it does not have a file list index. This file is used to indicate what files are in an mpq store, and all the other mpqs have this. 
- To get around this issue, we are currently combining the file lists of all other mpqs and using them as a base for the patch file list. This will find any replaced files.
- We need a mechanism for finding new files, as those currently will be unreachable.

**IsImploded**

- We needed to support the IsImploded flag in order to load the patch mpq data properly. This flag corresponds to a pkware implosion/explosion compression. 
- MPQStream had to be modified to support this case. 
- It is unknown if any other files in other mpqs are using this flag, we should be wary for side-effects of this change.

**CharStats**

- Revised because the patch version is different in its structure.
- The patch version includes PerLevelStatPoints, which is how many stat points a character gains to spend per level.
- Also: StartingSkills, which is a list (by string names, not ids) of what skills a character starts with.
- Also: BonusStrings, which are used by the UI to describe bonuses to certain skills
- See https://d2mods.info/forum/kb/viewarticle?a=389

**Level Details**

- Revised because the patch version is different in structure.
- QuestFlag: indicates whether a character can take a portal to this location
- Difficulties; some values in this object have been split into a separate difficulties object. The LevelDetails object now has a dictionary<eDifficulty, LevelDetailDifficulty> which can be used to figure out this info based on your current difficulty. This includes: monster level, monster density, size x, size y, monster spawns.
- Teleport: whether players can teleport in this level.
- Critter slots reduced to 4 (from 5)
- Monster slots reduced to 10 (from 25) (NOTE: sources claim that the d2 engine still accepts up to 25 if you manually add the columns back in; currently we do not support this and should consider expanding this functionality to support that case)
- See https://d2mods.info/forum/kb/viewarticle?a=301

**Todo**

- This check-in includes currently unused work related to reading missiles.txt. This will be completed at a later date and is currently not being loaded.
- Monstats needs to be revised. Currently it is disabled. The post-patch file has different columns.
- We need a mechanism for finding new files in the patch, such as monstats2.txt


nicholasdechiara:
Note: because eHero has None added, I had to make changes to SelectHeroClass in order to skip over None when trying to render / update rendering of hero classes.

nicholasdechiara:
Note: Armor, Misc, and Weapon had their columns move as well in the patch. Keep an eye out for other issues that may arise.

essial:
Thank you sir.

PR #64, opened 2019-01-05
Fix crash on exiting "select hero class" screen

cardoso:


cardoso:
@essial Changes done, mate :)

PR #82, opened 2019-11-04
Update .gitignore

liberodark:
Better to no add .exe for cross plateform project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 **/Client
 .idea/**/*
 .vs/**/*
+/OpenDiablo2.exe


liberodark:
Probably add in gitignore : rh.exe would be great better for no expand size

essial:
Rh.exe is needed for CI build automation to set windows release app icon.

liberodark:
yes but OpenDiablo.exe is not necessary for CI

essial:
Yeah, I agree with the OpenDiablo2.exe part, that was an issue we found last night. Just making sure you know we DO need rh.exe.

PR #107, opened 2019-11-08
common: use stringer and string2enum tools to convert back and forth …

mewmew:
…between enum and string


essial:
As with the other commit, I'll have to review this tomorrow before merging. Thanks!

PR #124, opened 2019-11-09
Vsync and FPS display toggles.

Ziemas:
They now toggle more reliably as well as work globally.

F6 toggle FPS display, F8 toggles vsync. Chosen with utmost care and not at all at random.

Check if you're happy with where i stuck the code for these first.

### DIFF
--- a/core/Engine.go
+++ b/core/Engine.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"path"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


core/Engine.go
```diff
@@ -173,6 +177,18 @@ func (v *Engine) Update() {
 		v.fullscreenKey = false
 	}
 
+	if inpututil.IsKeyJustPressed(ebiten.KeyF6) {
+		v.showFPS = !v.showFPS
+	}
+
+	if inpututil.IsKeyJustPressed(ebiten.KeyF8) {
+		if ebiten.IsVsyncEnabled() == true {
```
essial: Should be ebiten.SetVsyncEnabled(!ebiten.IsVsyncEnabled())

PR #125, opened 2019-11-09
Check if MPQ file was loaded before reading.

Ziemas:
And if it wasn't then tell the user which one instead of crashing.

### DIFF
--- a/core/Engine.go
+++ b/core/Engine.go
@@ -105,6 +105,9 @@ func (v *Engine) LoadFile(fileName string) []byte {
 	}
 	for _, mpqFile := range v.Settings.MpqLoadOrder {
 		archive, _ := mpq.Load(path.Join(v.Settings.MpqPath, mpqFile))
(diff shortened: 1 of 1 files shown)


essial:
Thank you sir. Many people will appreciate that error message!

core/Engine.go
```diff
@@ -105,6 +105,9 @@ func (v *Engine) LoadFile(fileName string) []byte {
 	}
 	for _, mpqFile := range v.Settings.MpqLoadOrder {
 		archive, _ := mpq.Load(path.Join(v.Settings.MpqPath, mpqFile))
+		if archive == nil {
```
mewmew: Another option would be to check the returned error, instead of discarding it to `_`.

```go
archive, err := mpq.Load(...)
if err != nil {
   log.Fatalf("Failed to load specified MPQ file: %s; %v", mpqFile, err)
}
```

PR #128, opened 2019-11-10
Restructured everything.

essial:
This PR restructures pretty much everything. The idea is to move files into a less insane package structure so people can figure out what is where without knowing the entire codebase.


vampy:
Hi, cool changes, but in my opinion the new package names with the `d2` prefix is just noise and goes against go package naming scheme (see https://blog.golang.org/package-names).

So when you import a package, instead of:
`github.com/OpenDiablo2/OpenDiablo2/d2common/d2enum`

it would be something like:
`github.com/OpenDiablo2/OpenDiablo2/common/enum`

averrin:
Folders in `d2data` have very obscure naming. Structures like dt1, d2dsst and etc...

mewmew:
> Folders in d2data have very obscure naming. Structures like dt1, d2dsst and etc...

I think `dt1` is a file format, so it makes sense to call it so, similar to [images/png](https://godoc.org/image/png) of the Go standard library.

mewmew:
Hi @vampy!

Glad to see so many new people join the project!

> it would be something like:
> `github.com/OpenDiablo2/OpenDiablo2/common/enum`

@vampy Naming packages is difficult, and even interpreting best practices of Go may be so too. Take the `common` package for instance.

From the blog post you linked, they mention:

> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. 

So, adding at least a `d2` prefix to `common` does help clarify the situation. Similar to using `stringutil` instead of just `util`.

While the package names are not perfect, it's a good start towards something more Go idiomatic, than what was before which was closer to how you'd structure projects in C#.

There may still be work to do, to break out core functionality into dedicated packages that each deal with a very specific task, but I think @essial did a great job at getting the ball rolling.

Also, to prevent too much stir of the Git repo, large scale changes like these should be made not too often.

That being said, if there are e.g. specific concepts of `d2common` that belong better in their own more descriptive package, then feel free to break them out. So long as the packages don't become too small. You know, finding the right fit, the balance, the sweet spot :)

Cheers,
Robin

PR #129, opened 2019-11-10
Improve run.sh

q3cpma:
* POSIX sh
* simpler distro detection
* don't create functions for stuff used only once
* add Gentoo (portage) support
* handle other distros gracefully
* ask before stealthily installing Go
* compile even when Go is already installed


essial:
Looks good, thanks!

PR #135, opened 2019-11-10
Simplify configuration + macOS support

AlexanderEkdahl:
This simplifies the steps to build from source into:

```
go install github.com/OpenDiablo2/OpenDiablo2
OpenDiablo2
```

No `config.json` required(but can still be provided) and operating specific paths are inferred from `runtime.GOOS`.

I also ran `go mod tidy` thereby removing unused dependencies.


.gitignore
```diff
@@ -5,3 +5,4 @@
 .idea/**/*
 .vs/**/*
 /OpenDiablo2.exe
+/OpenDiablo2
```
AlexanderEkdahl: Output from `go build` on unix based systems.

d2common/Configuration.go
```diff
@@ -2,54 +2,129 @@ package d2common
...
-	MpqLoadOrder    []string
-	SfxVolume       float64
-	BgmVolume       float64
+	Language          string
+	FullScreen        bool
+	Scale             float64
+	PauseInBackground bool
+	TicksPerSecond    int
+	FpsCap            int
+	VsyncDisabled     bool
```
AlexanderEkdahl: I had to switch these so that it would work with Go default values.

README.md
```diff
@@ -17,17 +17,14 @@ Diablo 2 and its content is ©2000 Blizzard Entertainment, Inc. All rights reser
...
 
-On windows this folder will most likely be in `C:\users\(you)\go\src\github.com\essial\OpenDiablo2`
-
-In the root folder, run `go get -d` to pull down all dependencies.
-
-To run the project, run `go run .` from the root folder.
-
-You can also open the root folder in VSCode. Make sure you have the `ms-vscode.go` plugin installed.
+```
+go install github.com/OpenDiablo2/OpenDiablo2
```
mewmew: Either use `go get` or add a `git clone` command above to retrieve the source code repository.

config.json
```diff
@@ -1,24 +0,0 @@
-{
```
mewmew: Rather than removing `config.json`, perhaps rename it to `config.json.default`?

I don't have Diablo II installed on my system, so trying to locate the MPQ archives in the *standard location* would fail. As such, I would use a custom `config.json` to run OpenDiablo2, and having an example config file is always useful.

d2common/Configuration.go
mewmew: That's good. Making the zero-value the default value.

README.md
AlexanderEkdahl: `go install` actually puts the code into your `$GOPATH` the same way a clone would.

AlexanderEkdahl: Or am I mistaken?

config.json
AlexanderEkdahl: Ah I agree. I will restore.

essial:
I've noticed your path names are capitalized, please make sure you sync with the latest version of the repo (we fixed all the filenames to be all lowercased). If you're on windows you may have to re-pull if it doesn't auto-rename the values.

README.md
mewmew: It does, I was mainly referring to the clone of the repo. So if you want to get started with OpenDiablo2, then you don't have the repository cloned yet to your local machine. `go get` can clone the repo, or we can clone it; e.g.

```
git clone https://github.com/OpenDiablo2/OpenDiablo2
cd OpenDiablo2
go install
```

That would work

Or we can run
```
go get github.com/OpenDiablo2/OpenDiablo2
```

Thinking about it, the first one is probably preferred, as we use Go modules.

corylanou: Because of modules now, I actually prefer the `git clone` as I want to put the project somewhere I'm going to `work` on it.  If you use `go get`, it will put it in your packages and not likely the working copy you intend to edit.  We should also be encouraging people to do the fork as well, and then update instructions for them to clone their fork instead.  Thoughts?

d2common/Configuration.go
```diff
@@ -2,54 +2,129 @@ package d2common
...
+	"Diablo II Expansion Movies",
+	"Diablo II Game Data",
+	"Diablo II Speech",
+	"Diablo II Music",
+	"Diablo II Sounds",
+	"Diablo II Movies",
+	"Diablo II Speech",
+}
+
+func LoadConfiguration() (*Configuration, error) {
```
corylanou: For a future revision, this should take a `os.Reader` so that we can easily write a test for this function.  Regardless, we could still write a test that uses the `testdata` directory to load different configs and make sure the defaults apply where necessary.

README.md
mewmew: Yeah, lets go with Git clone.

I don't think we need anything specific for fork instructions.

Just add something like this:

```
git clone https://github.com/OpenDiablo2/OpenDiablo2
git remote add myusername git@github.com:myusername/OpenDiablo2.git
```

Then, build like normal:

```
cd OpenDiablo2
go install
```

And when you make changes, pull to your own fork.

```
git checkout -b feature-foo
git commit -a -m 'add feature foo'
git push myusername feature-foo
```

And then submit a PR as normal.

essial:
This commit is now related to #174

AlexanderEkdahl:
All good comments and I am planning on working more on this and other tickets as soon as I get my development environment working again. I made the mistake of upgrading to macOS Catalina and it breaks the `glfw` dependency...

AlexanderEkdahl:
```
go build
# github.com/go-gl/glfw/v3.2/glfw
In file included from ../go/pkg/mod/github.com/go-gl/glfw@v0.0.0-20190409004039-e6da0acd62b1/v3.2/glfw/c_glfw_darwin.go:8:
../go/pkg/mod/github.com/go-gl/glfw@v0.0.0-20190409004039-e6da0acd62b1/v3.2/glfw/glfw/src/cocoa_window.m:989:9: warning: multiple methods named 'center' found [-Wobjc-multiple-method-names]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/AppKit.framework/Headers/NSCollectionViewCompositionalLayout.h:601:19: note: using
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/AppKit.framework/Headers/NSWindow.h:312:1: note: also found
```

On a related note: Diablo 2 refuses to run at all on macOS Catalina(they dropped 32-bit support). Another reason for OpenDiablo2 to exist!

d2common/Configuration.go
```diff
@@ -2,54 +2,129 @@ package d2common
...
+	if config.Scale <= 0 {
+		config.Scale = 1.0
+	}
+
+	if config.TicksPerSecond <= 0 {
+		config.TicksPerSecond = -1
+	}
+
+	if config.SfxVolume <= 0 {
+		config.SfxVolume = 1.0
```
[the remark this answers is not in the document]
AlexanderEkdahl: Good catch! I will fix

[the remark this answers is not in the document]
AlexanderEkdahl: Actually this is not possible as there is no way for Go to differentiate between the zero value(0.0) and the absence of the value being configured. 

I could add another boolean `mute`. Would that be better?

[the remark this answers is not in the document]
essial: Mute boolean is fine, may make more sense that way.

AlexanderEkdahl:
@essial Updated this PR

@mewmew Updated installation instructions. I agree with your workflow for working off a development branch.

AlexanderEkdahl:
@essial updated with more explicit error message. There is one remaining mpq file that I don't know if is even required or if something is missing. However, for now all of the assets are there that are required for running the game in its current state.

essial:
This PR is pending the addition of the ommited d2chars file as it is necessary for rendering the heros.

AlexanderEkdahl:
@essial I no longer have access to a macOS pre Catalina installation so I am unable to find the missing MPQ file.

PR #138, opened 2019-11-10
Fix checkboxes

Ziemas:
And change the names of things because it was confusing me.

### DIFF
--- a/d2render/d2ui/checkbox.go
+++ b/d2render/d2ui/checkbox.go
@@ -11,7 +11,7 @@ import (
 
 type Checkbox struct {
 	x, y          int
-	pressed       bool
+	checkState    bool
 	visible       bool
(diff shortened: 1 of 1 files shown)


d2render/d2ui/checkbox.go
```diff
@@ -46,7 +46,7 @@ func (v Checkbox) Draw(target *ebiten.Image) {
 		Filter:        ebiten.FilterNearest,
 	}
 	opts.GeoM.Translate(float64(v.x), float64(v.y))
-	if v.pressed == false {
+	if v.checkState == false {
```
mewmew: I know this is merged already, but a suggestion for booleans:

```diff
-if v.checkState == false {
+if !v.checkState {
```

The same goes for `v == true`

PR #143, opened 2019-11-12
Added more functionality to character select screen.

essial:
This PR adds a lot of functionality to the new char screen, adds more game state stuff, and gets into 'the game', albeit as a black screen. There's more work to be done on the game state object as well as the character select screen, but this is a step in the right direction.
 
Known issues:
- Scrolling isn't fully implemented yet
- Sometimes the last character doesn't get deleted for some reason
- Loading can be slow as atlas logic hasn't been added to the character frame loader


ramiabdh:
لااعرف استخدمه ولا ادري ايش الفايده منه

PR #146, opened 2019-11-12
Fix Folder location

liberodark:


essial:
When I get home this afternoon I’ll figure out what to do about the base path, I’ll need to see what Mac users need this set to.

essial:
Looks like the base path should be based on os.UserConfigDir. Make that change and I'll accept the PR.

liberodark:
Yes but all configuration of software or games is in .config not is home is not great.

Ziemas:
`os.UserConfigDir` respects the XDG Base Directory specification and will use `$XDG_CONFIG_HOME` or `~/.config/` if that variable is unspecified.

We can probably use it for other platforms as well as per follows

> On Darwin, it returns $HOME/Library/Application Support. On Windows, it returns %AppData%. On Plan 9, it returns $home/lib.

AlexanderEkdahl:
+1 for `UserConfigDir`. Works well on macOS.

PR #159, opened 2019-11-13
Additional credits in the credits screen #154

### LINKED ISSUE #154: Additional credits in the credits screen
We need a string list added to the credits screen where we can easily add credits for OpenDiablo 2 contributors. All this needs to do is prepend this string list to the strings we load from the actual credits file.

averrin:
fixes #154


essial:
Thank you!

PR #161, opened 2019-11-13
Idle animation and hover gets out of sync

TehGoat:
When selecting a hero the idle and hover animation gets out of sync


essial:
Nice catch, thanks again!

PR #169, opened 2019-11-14
Render in two passes

Ziemas:
To avoid drawing over objects.

See if you like it I guess. I don't really know what I'm doing, especially with pointers and stuff.

### DIFF
--- a/d2render/d2mapengine/engine.go
+++ b/d2render/d2mapengine/engine.go
@@ -130,12 +130,12 @@ func (v *Engine) GenTiles(region *EngineRegion) {
 }
 
 func (v *Engine) RenderRegion(region EngineRegion, target *ebiten.Image) {
-	var tilesToRender []RegionTile
+	tilesToRender := make([]RegionTile, len(region.Tiles))
 
-	for _, tile := range region.Tiles {
+	for i, tile := range region.Tiles {
 		sx, sy := d2helper.IsoToScreen(tile.tileX+region.Rect.Left, tile.tileY+region.Rect.Top, int(v.OffsetX), int(v.OffsetY))
 		if sx > -160 && sy > -160 && sx <= 880 && sy <= 1000 {
-			tilesToRender = append(tilesToRender, *tile)
+			tilesToRender[i] = *tile
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


d2render/d2mapengine/engine.go
```diff
@@ -86,52 +96,56 @@ func (v *Engine) GetRegionAt(x, y int) *EngineRegion {
...
 	}
 }
 
-func (v *Engine) RenderTile(region *Region, offX, offY, x, y int, target *ebiten.Image) {
-	tile := region.DS1.Tiles[y][x]
-	for i := range tile.Floors {
-		if tile.Floors[i].Hidden || tile.Floors[i].Prop1 == 0 {
-			continue
+func (v *Engine) RenderRegion(region EngineRegion, target *ebiten.Image) {
+	var tilesToRender []RegionTile
```
corylanou: A micro optimization here would be to:
```suggestion
	tilesToRender := make([]RegionTile, len(region.Tiles))
```

This will allow to allocate all the memory at once.  Append has the potential to create a lot of garbage otherwise.

d2render/d2mapengine/engine.go
```diff
@@ -86,52 +96,56 @@ func (v *Engine) GetRegionAt(x, y int) *EngineRegion {
...
+
+	for _, tile := range region.Tiles {
+		sx, sy := d2helper.IsoToScreen(tile.tileX+region.Rect.Left, tile.tileY+region.Rect.Top, int(v.OffsetX), int(v.OffsetY))
+		if sx > -160 && sy > -160 && sx <= 880 && sy <= 1000 {
+			tilesToRender = append(tilesToRender, *tile)
```
corylanou: You can change this code to match the above suggestion now:

```go
for i, tile := range region.Tiles {
		sx, sy := d2helper.IsoToScreen(tile.tileX+region.Rect.Left, tile.tileY+region.Rect.Top, int(v.OffsetX), int(v.OffsetY))
		if sx > -160 && sy > -160 && sx <= 880 && sy <= 1000 {
			tilesToRender[i] =*tile
```

essial:
I agree with the suggestions @corylanou made. Once these adjustments are made we will accept the PR. Thanks @Ziemas !

Ziemas:
Thanks for the suggestions! Applied those and rebased on master.

PR #172, opened 2019-11-14
Toggle tile display in map test with F7

Ziemas:


d2core/d2scene/map_engine_testing.go
```diff
@@ -10,13 +10,15 @@ import (
 
 	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2interface"
 
+	"os"
```
mewmew: @Ziemas, try to format this source file using `goimports`. That should group related packages together. In this case the `os` package would be placed alongside the other packages of the standard Go library.

Run:
```
go get golang.org/x/tools/cmd/goimports
goimports -w d2core/d2scene/map_engine_testing.go
```

essial:
@Ziemas if you use golang you can add file watchers in the settings menu that'l automatically do this. VS code should auto do this as well if you use the go plugin. Either way, once the imports are re-ordered I'll merge this PR as well. Thanks again!

Ziemas:
goimports just leaves it where it is for me, and usually vscode organizes them automatically. Dunno why it's ending up in the wrong place.

I'll move it manually for now.

Ziemas:
Added sub-tile display, cycle through modes with F7.

PR #173, opened 2019-11-14
Switcher with presets & better tiles randomizing

averrin:
1. Better getTile function (with some dance around cacheIndex)
2. You can switch region and cycle presets without fear. Only Act 5 seems broken. Idk, its wrong table, broken code or broken me.


essial:
Great, can't wait to see the final PR!

PR #178, opened 2019-11-15
Closes Issue#176

Myles-Glass:
"Fatal is equivalent to Print() followed by a call to os.Exit(1)." This is why it is crashing.


essial:
Wouldn't this crash on `scanner := bufio.NewScanner(file)` due to file being nil? I believe we should return an empty array after printing the error.

PR #182, opened 2019-11-15
Adjust sub-tile rendering offset.

Ziemas:
-5 Y when rendering based on sub-tile matches real D2 (at least for
rendering act 1 stash)


essial:
Epic PR, would merge again. +++

Issue #77, opened 2019-11-03
Credits screen line spacing wrong


essial:
The spacing between the lines in the credits screen does not match the official D2 client. The spacing needs to be adjusted to accurately match what the original client does.

Tello-:
Looks like both the spacing between lines AND the justification of each line needs adjustment. I opened up the credit sequence on DII and held my cursor in the middle and all the lines are centered with the header.

kazgaroth011:
Are you sure? I made adjustments by comparing with original D2, so the spacing between lines is now exactly 19, 2*19 or 3*19 pixels, just as in the original. Also the text should be centered at 400 pixel width. The only thing off is text color, which I can't get right.

Tello-:
I realized that I had put it in windowed mode when I did the alignment thing with the cursor. I didn't realize this until I was trying to fall asleep haha. I don't know if there is ever a reason or feature to play in windowed... so possibly disregard.

marti1125:
Hi @essial  I would like to contribute with this

essial:
Sure, and I also suggest hanging out in our discord channel as well!

essial:
Closing this for now, not a big issue and I don't want to commit resources to it.

PR #183, opened 2019-11-15
Some corrections for credits

kazgaroth011:
Related to issue #77


kazgaroth011:
Note that the text color is off (I couldn't match it, text gets blurred and RGB colors don't match those in the code). The heading especially needs to be less bright,

essial:
Thanks!

PR #184, opened 2019-11-16
Parallel scene loading

AlexanderEkdahl:
Loads scenes much faster ⏩ Not sure if this is a good idea or not....

Previously it loaded one step for every tick of the update loop. This assume that the functions returned by `Load()` can be executed concurrently.

### DIFF
--- a/d2core/engine.go
+++ b/d2core/engine.go
@@ -8,29 +8,20 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
-	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2resource"
-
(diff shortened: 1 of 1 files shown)


essial:
We cannot run ebiten image calls across a render boundary or it will randomly produce errors. The fix will be a bit more complex.

AlexanderEkdahl:
That makes sense. Is there another way we can separate the assets required for a scene from the rendering?

essial:
We've had many discussions about this and other things on the discord channel, and the hope is that very soon the ebiten maintainer will add the concept of image lists. He just added PBOs which made some users go from 8 minute loads to 8 seconds. If image lists are implemented, the texture atlas thing goes away, and we will pump ebiten with a bunch of pixel buffer arrays, which SHOULD be possible across frame boundaries as they are queued calls.

PR #192, opened 2019-11-17
Update Ebiten Reference

ndechiara:
The new Ebiten version offers significantly improved map test load times.


essial:
Thanks, this makes a huge difference for people who have had 8 MINUTE load times on maps. Thanks ebiten team!

PR #200, opened 2019-11-17
Create tiles cache on region loading

### LINKED ISSUE #199: Latest batch of commits lead to poor performance
With the latest batch of commits from today(and probably late yesterday), I get poor performance on map test. When scrolling the map with arrow keys, I get FPS of 5,6,7, wheres on static map it goes to 150-300 depending on the scene.

averrin:
fixes #199 
Slightly longer on load, but smooth at rendering


Dragomir-Ivanov:
This fixes poor scrolling performance, however now it takes longer to load a level. Tristam test map, for example takes ~1 minute to load from 5400rpm HDD on a laptop. During this load time OD2 window even becomes gray on Ubuntu, indicating that the app is not responsive for the moment.

In any case, good work guys, maybe in the coming months I will be able to give a helping hand to this project! :+1:

essial:
We're working with the ebiten dev on this. We resolved an issue with his atlas logic causing hundreds and hundreds of calls. Replace the ebiten line in go.mod with the following:
`github.com/hajimehoshi/ebiten v1.11.0-alpha.0.20191117152313-63f9ac2ccc2a`

There's another issue behind that we're still working through. This is something we're slowly improving over time of course, so I appreciate your patience 😊

essial:
Ok, so I've tracked the final issue to the fact that it caches each and every wall, floor, and shadow tile individually, even if they're exactly the same as another tile. So it's creating almost 7,000 tile cache images at once. It's a miracle of ebiten's performance improvements that it runs at ALL. I'll fix this issue today as soon as I can!

Issue #197, opened 2019-11-17
NPC and Object loaders are trash


essial:
The NPC and object loading code (that creates the image caches and such) are extremely slow and dumb. It causes massive lag on the map screens. If no one picks this up, I will fix this next week. I am putting this issue in to save everyone else the time from reporting it.

Basically right now it renders every individual component for every frame, creating hundreds of objects for no reason. What it should do is create a frameset for all animations for a direction, and render the components directly on that frame so that each render is just a single frame render call.

averrin:
Do you suggest creating one big image from once rendered tiles? At least floor tiles (walls can be transparent, iirc)

essial:
My thought is to pre-generate all of the frames in a set of threads, basically create the pixel byte array in memory, but don't push it to video. Then on render simply setpixels to the new frame data.

averrin:
So, we needn't have individual cache image per every tile, we should just create pixels for every possible in this region tile variant. And then just create a reference to it.

Dragomir-Ivanov:
Guys, there is a resource leak somewhere with the latest OD2. After some cycling on the test maps, my OD2 process end up taking 6GB of memory.

essial:
Try now.

essial:
I believe this has been resolved a while back.

PR #203, opened 2019-11-17
More map engine improvements.

essial:
Made a few more minor improvements to the object loader and map rendering.


essial:
Fixes #197 and #199

PR #210, opened 2019-11-19
Avoid linux case sensitivity issues in mpq loading

Ziemas:


essial:
This actually needs to be in D2Common so that the toolset can also take advantage of the new logic.

Ziemas:
Submitted to d2shared instead.

PR #222, opened 2019-11-22
Fix CJK render problem partially

0b11111111:
This fix contains a character converting table, which is used in CJK family in order to map Unicode characters to frames index.

HOWEVER, THIS FIX WILL IMPACT THE PERFORMANCE ISSUE HEAVILY, because CJK fonts are normally huge. What I trying to do is to use a `LazySprite` to render what it needs, but that hits the core of the engine, which I'm not very familiar :(

See also OpenDiablo2/D2Shared#20


0b11111111:
Oh I forget the screenshot of my Chinese version, take it here:

d2cn

essial:
Are you in discord? I'd love to help fix this performance issue!

PR #226, opened 2019-11-22
Improve little performance of rendering CJK

krnl0x0:
Added a strategy to render the needed fonts images lazily/manually. It seems work, at least better.


nicholas-eden:
I broke this PR by moving the loading of dc6 files to D2Shared.  The loader doesn't do the actual rendering of the frames until the rendered frame is requested so this could be modified to call `Render()`.

essial:
Closing this for now as there's a lot of core engine changes and reshuffling going on. After the codebase stabilizes we'll re-visit this.

PR #234, opened 2019-11-24
Extract sprite from dc6 from D2Shared

nicholas-eden:
Remove processing of dc6 file type from sprite creation.  Use dc6 loader from D2Shared.

Depends on https://github.com/OpenDiablo2/D2Shared/pull/25


essial:
I just merged the D2Common changes. You'l need to update the D2Common ref on this project before it will pass the checker. Just replace the version with 'latest' and save. If you're using GoLand it'l auto update to the latest, otherwise run go get or something.

nicholas-eden:
Thanks, I've updated go.mod.

PR #240, opened 2019-11-29
Add basic movement

j0y:
This pull request adds ability to move around.
I added 2 new variables to `AnimatedEntity` to keep track of target location. Character velocity right now hardcoded in `StepToTarget()`.
Code also depends on `d2helper.AlmostEqual` function to compare floats
`
func AlmostEqual(a, b, threshhold float64) bool {
    return math.Abs(a - b) <= threshhold
}
`
Tell me if this code gonna be usefull and I'll make pull request for D2Shared.

### DIFF
--- a/d2core/d2scene/character_select.go
+++ b/d2core/d2scene/character_select.go
@@ -226,7 +226,7 @@ func (v *CharacterSelect) Render(screen *ebiten.Image) {
 		v.characterNameLabel[i].Draw(screen)
 		v.characterStatsLabel[i].Draw(screen)
 		v.characterExpLabel[i].Draw(screen)
-		v.characterImage[i].Render(screen, v.characterNameLabel[i].X-40, v.characterNameLabel[i].Y+50)
+		v.characterImage[i].Render(screen, float64(v.characterNameLabel[i].X-40), float64(v.characterNameLabel[i].Y+50))
 	}
 	if v.showDeleteConfirmation {
 		ebitenutil.DrawRect(screen, 0.0, 0.0, 800.0, 600.0, color.RGBA{0, 0, 0, 128})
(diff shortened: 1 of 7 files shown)


essial:
Looks good. Do note that eventually there is going to be a list of coordinates to follow (that will be generated by A* pathfinding). My only comment is that we need to take time into account, and also make sure the character doesn't overshoot the target (if distance < how far we could move, move to the target immediately, then re-call step with the remaining time after how long it would take to move).

j0y:
Redid step calculation based on tick time, made the character teleport to target if it's within one step.

essial:
Looks like CI is randomly failing for no reason now.. I'll check it out in the AM.

Ziemas:
Nah, CI is failing because of the proposed change to d2shared missing.

essial:
Yeah it's annoying. It's failing to get libudev... Anyway go ahead and do the PR for D2Shared first so you can check this in with updated refs.

j0y:
Added changes to `Render` and `IsoToScreen` calls to remove float to int conversions. I am not sure about this code: https://github.com/OpenDiablo2/OpenDiablo2/pull/240/files#diff-67e2b6d34afa4e9c4bce79fe34eebf22R203 because I don't know when the Hero is supposed to rendered (before or after walls?) and how constants 400, 300 can be replaced with meaningful variables.

Ziemas:
The hero should be drawn at the same time as objects and npc's, before upper walls.

The way things are currently set up LocationX and LocationY is the current tile of the entity, within that tile there are Sub-tiles (also called sub-cell's in the codebase). Pathing and movement is based on these sub-tiles as far as I know.

j0y:
Thanks, that's good to know! I didn't change rendering order, just moved it out of the loop. As I understand `Render()` must be called with screen coordinates, so we can use `v.Hero.Render(target, 400, 300)` to always have him in the center of the screen (assuming 800×600 resolution)?

Ziemas:
Right now you're rendering the hero after all walls have been rendered, if you do it inside RenderPass2 only when for the tile that hero is actually in the hero will naturally render in front of the walls of the current tile but behind the walls of the tile below.

So something like this in RenderPass2
```go
if int(math.Floor(v.Hero.AnimatedEntity.LocationX)) == x && int(math.Floor(v.Hero.AnimatedEntity.LocationY)) == y {
	if v.Hero != nil {
		v.Hero.Render(target, 400, 300)
	}
}
```

Issue #104, opened 2019-11-07
Map rendering loop


essial:
**Is your feature request related to a problem? Please describe.**
Currently the map engine runs through every tile of every region on the map render loop. It should simply run through x/y tiles once, for the tiles 'visible' on the screen (keeping in mind the verticality of some tiles on the Y axis).

**Describe the solution you'd like**
We need to do one x/y render loop that does a square around the viewport and picks tiles from the last region in the region list that has a tile at that location.

**Additional context**
As we add more and more regions to map generation, the rendering is going to become incredibly slow until this change is made.

PR #246, opened 2019-12-09
Create viewport and camera types; simplify coordinate systems

FooSoft:
Hi there, this seems like a pretty cool project; I've decided it would be cool to contribute to it. I was actually working on my own Diablo 2 engine implementation ([Lazarus](https://github.com/FooSoft/lazarus)), but this project is much further along and has better community traction.

Anyway, I was going to do #104 but as it stands, the coordinate systems in the project are very confusing. I created viewport and camera constructs which should make it much simpler to support variable resolutions, spatial partitioning, custom UI in the future.

With this change, it is now simpler to work in the three coordinate systems (screen, world, and isometric). Once this is in, I will go to work on the map system; I think it can be massively simplified :smile:

### DIFF
--- a/d2core/d2scene/game.go
+++ b/d2core/d2scene/game.go
@@ -1,6 +1,8 @@
 package d2scene
 
 import (
+	"image/color"
+
 	"github.com/OpenDiablo2/D2Shared/d2common/d2enum"
 	"github.com/OpenDiablo2/D2Shared/d2common/d2interface"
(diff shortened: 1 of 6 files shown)


nicholas-eden:
Nice, I didn't like all those hard coded values.  This is a lot cleaner.

FooSoft:
Yeah, dealing with all of the offsets was kind of difficult as well (not to mention not always being clear on what units something is in).

essial:
Glad to have you on board, this simplifies things a lot!

PR #253, opened 2019-12-16
Add asset manager

FooSoft:
Handles in-memory caching of frequently used files, MPQ archives. Memory budgets can be set to determine when files get evicted from the cache and have to be reloaded from disk.


FooSoft:
Now relies on D2Common MPQ change; will update vendoring when that goes in.

PR #257, opened 2019-12-18
Configuration cleanup

### LINKED ISSUE #174: Generate configuration on first run
Currently the configuration file works for windows users by default, and for linux users who installed the game via wine. This means that mac users do not have a functional experience out of the box.

When the engine starts, it should determine if the config file exists. If it doesn't, it should detect the OS type, and generate a default configuration file based on this host.

FooSoft:
* Do not require configuration file; one is created with sensible defaults for current OS
* Delete existing configuration from git, it no longer needs to be checked in

Fixes #174


essial:
You're going to make a LOT of people happy with this one. Great work!

PR #295, opened 2020-02-04
Move hero by holding down mouse button

j0y:
Right now hero does not change direction when cursor is moved. This was working before commit https://github.com/OpenDiablo2/OpenDiablo2/commit/8de0cde818d0f536d9b0de7712137f8b39c8c1c4 so I made this small fix.

### DIFF
--- a/d2game/d2player/game_controls.go
+++ b/d2game/d2player/game_controls.go
@@ -47,18 +47,29 @@ func (g *GameControls) OnKeyDown(event d2input.KeyEvent) bool {
 
 func (g *GameControls) OnMouseButtonDown(event d2input.MouseEvent) bool {
 	if event.Button == d2input.MouseButtonLeft {
-		px, py := g.mapEngine.ScreenToWorld(event.X, event.Y)
-		px = float64(int(px * 10)) / 10.0
-		py = float64(int(py * 10)) / 10.0
-		heroPosX := g.hero.AnimatedEntity.LocationX / 5.0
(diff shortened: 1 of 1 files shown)


essial:
I'm not sure this was in the original game, can someone else confirm as well?

essial:
So I talked with @FooSoft and I think this isn't exactly what's needed. Movement is currently calculated when you click, but what we should also do is when the mouse is held down, every X amount of time it should re-path. I can't really tell by playing around with the official game, but it looks like it's at a rate of like 1/4th a second (250ms) or so.

noobmagnet:
> I'm not sure this was in the original game, can someone else confirm as well?

Yes, this is in the original game

essial:
> > I'm not sure this was in the original game, can someone else confirm as well?
> 
> Yes, this is in the original game

I should have clarified. It appears to be running every frame, and I'm not sure we should pathfind every single frame, could get a bit heavy on the client. Even if it DOES pathfinder every frame on the original's game, that's only a rate of 25 per second, so that needs to be accounted for before re-pathing.

noobmagnet:
> > > I'm not sure this was in the original game, can someone else confirm as well?
> > 
> > 
> > Yes, this is in the original game
> 
> I should have clarified. It appears to be running every frame, and I'm not sure we should pathfind every single frame, could get a bit heavy on the client. Even if it DOES pathfinder every frame on the original's game, that's only a rate of 25 per second, so that needs to be accounted for before re-pathing.

Oh I got yuh my bad, you were on another level lol

PR #305, opened 2020-02-18
removed ebiten references from d2input.go and input_manager.go

dknuth0101:
moved all ebiten references into ~/d2core/d2input/ebetininput.go

### DIFF
--- a/d2core/d2input/d2input.go
+++ b/d2core/d2input/d2input.go
@@ -2,13 +2,17 @@ package d2input
 
 import (
 	"errors"
-
-	"github.com/hajimehoshi/ebiten"
 )
 
 var (
 	ErrHasReg = errors.New("input system already has provided handler")
 	ErrNotReg = errors.New("input system does not have provided handler")
+)
+
+type InputBackend int
+
+const (
+	Ebiten InputBackend = iota
 )
 
 type Priority int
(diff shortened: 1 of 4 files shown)


dknuth0101:
closed because I thought I had removed `buttonMod` and `keyMod` from the inputManager struct, but i didn't.

d2core/d2input/d2input.go
```diff
@@ -207,6 +210,18 @@ type MouseMoveHandler interface {
 
 var singleton inputManager
 
+func Initialize(backend InputBackend) error {
+	singleton.key = make(map[Key]int)
+	singleton.mouseButton = make(map[MouseButton]int)
+
+	if backend == Ebiten {
```
FooSoft: Make this a switch, and return an `error` if there is no matching backend.

d2core/d2input/ebiteninput.go
```diff
@@ -0,0 +1,226 @@
...
+	im.key[KeyR]				= int(ebiten.KeyR)
+	im.key[KeyS]				= int(ebiten.KeyS)
+	im.key[KeyT]				= int(ebiten.KeyT)
+	im.key[KeyU]				= int(ebiten.KeyU)
+	im.key[KeyV]				= int(ebiten.KeyV)
+	im.key[KeyW]				= int(ebiten.KeyW)
+	im.key[KeyX]				= int(ebiten.KeyX)
+	im.key[KeyY]				= int(ebiten.KeyY)
+	im.key[KeyZ]				= int(ebiten.KeyZ)
+	im.key[KeyApostrophe]		= int(ebiten.KeyApostrophe)
```
FooSoft: Weird indentation?

d2core/d2input/ebiteninput.go
```diff
@@ -0,0 +1,226 @@
+package d2input
+
+import (
+	"github.com/hajimehoshi/ebiten"
+	"github.com/hajimehoshi/ebiten/inpututil"
+)
+
+func ebitenInput(im *inputManager) error {
+	im.key[Key0]				= int(ebiten.Key0)
```
FooSoft: Package is still `d2input`, so `ebiten` dependency is still here.

dknuth0101: was i supposed to make a new directory?

FooSoft: As I mentioned earlier, it doesn't matter how many or few files you have in a package. This change just moves the ebiten reference to a different file, and does not actually remove it altogether. You would have to create a new package underneath d2input (for example d2inputebiten), and put the ebiten bits in there.

d2core/d2input/ebiteninput.go
essial: Make sure your editor is honoring the .editorconfig file.

dknuth0101: it is, those were just logically distinct (keys vs mouse buttons), so i didn't feel like it needed the same indentation. I'll change it, though.

FooSoft:
Most of these changes are not actually necessary. What you are trying to accomplished regarding the constants can be done very easily without updating them at runtime or stuff like that.

1. d2input exports key and button constants like before, although they are not based on ebiten values.
2. d2input references d2inputebiten, or whatever it would be called, which has its own constants, based on ebiten values.
3. d2input has functions to convert between d2input constants and d2inputebiten constants.
4. d2input creates d2inputebiten instance and calls through to it, converting constants where necessary.
5. profit

FooSoft:
This change requires a fair amount of knowledge about how to properly do package and overall layer management in Go... I would honestly just do a smaller scale change first, and then come back to this once you are a bit more comfortable with how the language functions. It's not a very difficult language but it can be easier to point out conventions and best practices when starting with something smaller (especially if you are new to the language) 😸

dknuth0101:
This is literally the first code I've tried to write in Go, still learning and I want to figure this out.

PR #309, opened 2020-02-22
Direction changes.

Ziemas:
According to Necrolis the game uses 64 directions for entities which then get mapped down to however many directions a specific asset has by way of a lookup table.


d2core/d2asset/animation.go
```diff
@@ -251,12 +251,9 @@ func (a *Animation) GetDirectionCount() int {
 }
 
 func (a *Animation) SetDirection(directionIndex int) error {
-	if directionIndex >= len(a.directions) {
-		return errors.New("invalid direction index")
-	}
-
-	a.directionIndex = directionIndex
+	a.directionIndex = d2dcc.Dir64ToDcc(directionIndex, len(a.directions))
```
nicholas-eden: Should we remove the error return?  Or panic if directionIndex >= 64?

d2core/d2asset/composite.go
```diff
@@ -126,8 +127,12 @@ func (c *Composite) createMode(animationMode, weaponClass string, direction int)
 		return nil, err
 	}
 
-	if direction >= cof.NumberOfDirections {
-		return nil, errors.New("invalid direction")
+	// oh god how do i math
+	offset := (64 / cof.NumberOfDirections) / 2
+	entityDirection := int(math.Trunc((float64(direction+offset)-64.0)*(-float64(cof.NumberOfDirections)/-64.0) + float64(cof.NumberOfDirections)))
```
nicholas-eden: It looks like `math.Trunc` is redundant
```suggestion
	int((float64(direction+offset)-64.0)*(-float64(cof.NumberOfDirections)/-64.0) + float64(cof.NumberOfDirections))
```

d2core/d2map/animated_entity.go
```diff
@@ -42,21 +41,9 @@ func (ae AnimatedEntity) GetDirection() int {
...
-	case 8:
-		layerDirection = d2dcc.CofToDir8[ae.direction]
-	case 16:
-		layerDirection = d2dcc.CofToDir16[ae.direction]
-	case 32:
-		layerDirection = d2dcc.CofToDir32[ae.direction]
-	}
-
-	ae.animation.SetDirection(layerDirection)
+	ae.animation.SetDirection(ae.direction)
```
nicholas-eden: This is so much better

d2core/d2asset/animation.go
Ziemas: Added the check back with >= 64

PR #312, opened 2020-02-24
adding event bus, not currently used anywhere

dknuth0101:
Provides a global event bus, exports `On(string, func(payload ...interface{}))` and `Emit(string, payload ...interface{})`.

I started listing events, but didn't really finish it. Don't even know if that would be the right place to put them.


dknuth0101:
we'll decide on a better implementation in the future.

PR #319, opened 2020-02-25
Feature pl2 asset manager

dknuth0101:
cleaned up after feedback from previous PR

can now load pl2 files through `d2asset.loadPaletteTransform(path)`


dknuth0101:
Oops, some minor things need to be edited.

dknuth0101:
I should have mentioned that the PL2 structs came from Lectem's [Worldstone repo](https://github.com/Lectem/Worldstone/), I'm just copy/pasting. Lectem's code can encode PL2 files from a base palette (pal.dat) file, as well as decode existing PL2 files.

PR #323, opened 2020-03-01
refactored d2input, abstracted ebiten to backend interface

dknuth0101:
main.go is currently hardcoded to init input system with ebiten as backend


d2core/d2input/d2input_ebiten/backend_ebiten.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"github.com/OpenDiablo2/OpenDiablo2/d2core/d2input/keyboard"
+	"github.com/OpenDiablo2/OpenDiablo2/d2core/d2input/mouse"
+
+	"github.com/hajimehoshi/ebiten"
+	"github.com/hajimehoshi/ebiten/inpututil"
+)
+
+func keyToEbiten(k keyboard.Key) ebiten.Key {
+	return ebiten.Key(k)
```
FooSoft: Does this just not assume that all of the constants are the same between ebiten and the values that you added? How will this work when you try to add another input backend?

d2core/d2input/keyboard/keyboard.go
```diff
@@ -0,0 +1,209 @@
...
+	KeySlash
+	KeySpace
+	KeyTab
+	KeyUp
+	KeyAlt
+	KeyControl
+	KeyShift
+)
+
+var Keys = [...]Key{
```
FooSoft: Don't use variables for things that are actually supposed to be constants.

dknuth0101: I don't thing the array can be const. It's just there so we can range over the keys elsewhere. How should it be done as a const?

dknuth0101: Okay, I've addressed this in my recent commits.

d2core/d2input/d2input_ebiten/backend_ebiten.go
dknuth0101: A map is now explicitly defined, and not exported. It's now part of init for the input backend.

dknuth0101:
hmm... the midi driver breaks the build, the build environment needs to have libs for portmidi. I looked into using other midi drivers, but same issue.

PR #325, opened 2020-04-02
Fix and QA

liberodark:


liberodark:
image
Have 15 other issue for go to A+

PR #328, opened 2020-05-07
Item Affixes

dknuth0101:
Added resource paths for MagicPrefix.txt and MagicSuffix.txt to d2common/d2resource/resource_paths.go

Added calls in main.go to load the files

Added classes for item affixes and item affix groups in d2common/d2data/d2datadict/item_affix.go

Added constants for item affix super/sub types

Minor edit to d2common/data_dictionary.go as it was allocating more lines than necessary.


d2common/data_dictionary.go
```diff
@@ -20,16 +20,16 @@ func LoadDataDictionary(text string) *DataDictionary {
 	for i, fieldName := range fileNames {
 		result.FieldNameLookup[fieldName] = i
 	}
-	result.Data = make([][]string, len(lines)-1)
```
essial, line 23: Why did you make this change? Won't repeatedly calling append() add a bunch of stuff that the gc has to free after this is done?

nicholas-eden: You can at least allocate the full array and leave the slice empty.  But it looks like this is to fix an off by one error: it allocates len - 1 but since the loop started at index 1 it left an empty record.  Imo change it to len - 2 if we want to be kind to GC or `make([][]string, 0, len(lines) - 2` if you want to be defensive.

dknuth0101: Yes, it was to fix an OBO error. I've made the edit, thank you.

PR #344, opened 2020-06-20
Migrate VSCode Golang extension

dvdmuckle:
The VSCode Golang extension, formally as `ms-vscode.go`, has now been moved over to the Golang org and is now at `golang.go`.


essial:
Good catch, thanks!

PR #364, opened 2020-06-20
Feature/automap

danhale-git:
This is a loader for AutoMap.txt. It's based on other loaders in d2datadict.

https://d2mods.info/forum/kb/viewarticle?a=419

### DIFF
--- a/d2common/d2data/d2datadict/automap.go
+++ b/d2common/d2data/d2datadict/automap.go
@@ -2,13 +2,27 @@ package d2datadict
 
 import (
 	"log"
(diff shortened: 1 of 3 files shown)


danhale-git:
There is an "Expansion" row handled in gems.go, level_presets.go and level_types.go, as well as in automap.go. This row has all empty field values and seems to be a placeholder. Should we handle it in d2datadict.LoadLevelPresets?

essial:
I wouldn't worry about it if there is no data to load.

PR #368, opened 2020-06-21
Fixed patreon to Patreon

dirkpuge:
Fixed patreon to Patreon


essial:
Thanks, my spelling is the _worst_..

PR #369, opened 2020-06-21
camera offset for ui panels :

### LINKED ISSUE #358: Camera offset with open UI panels
When a panel is open, the camera should be offset so that the player is centered on the half that isn't covered.

Haashi:
closes #358 

added maprenderer viewport to be aligned left or right

calling alignement on keyPress in game_controls

I'm unsure of the change inside isOrthoRectVisible. Without this change, the viewport aligned on right would look like this : 
image

If you see room for improvements, let me know


essial:
Would it be possible to turn the left/right/normal into an enum so it isn't possible to accidentally call left/right multiple times?

PR #370, opened 2020-06-21
fix bug from PR#369

Haashi:
fix #369 :

-forgot to assign alignement
-defaultScreenRect instead of screenRect because of issue mentionned in original comment of #369


Ziemas:
You accidentally included your config.json and that go.mod change.

Haashi:
Thanks you really much.
I spent more time on fixing my git/github misstakes than on actual reading/testing/coding the feature ahah.

PR #374, opened 2020-06-21
Escape menu - Keyboard support and minor improvements

### LINKED ISSUE #353: Escape Menu support
We need to implement the escape menu on the game so that players can leave the game, adjust settings, etc. This is that big menu that comes up when you hit ESC during gameplay. For now, just being able to call it up and have stub functions executed if they select an option is good enough for this task.

Also, for single player games, this should cause the game to pause in the background.

aziule:
Hi, this is my first PR on the project so please don't hesitate to let me know if I broke the structure/design/whatever.

It partially fixes #353 (see comments there).

Work done:
* Add support for up/down/enter keys in the ingame "escape menu"
* Add sound when selecting an item
* Reset the menu when toggling it
* Prevent opening the menu over the inventory or hero stats panels
* I also took the liberty to remove some methods that were not part of an interface and were not used, so as to avoid confusion or unintended side effects (YAGNI)

### DIFF
--- a/d2common/d2resource/resource_paths.go
+++ b/d2common/d2resource/resource_paths.go
@@ -253,6 +253,7 @@ const (
 
 	// --- Sound Effects ---
 
+	SFXCursorSelect        = "cursor_select"
 	SFXButtonClick         = "cursor_button_click"
 	SFXAmazonDeselect      = "cursor_amazon_deselect"
 	SFXAmazonSelect        = "cursor_amazon_select"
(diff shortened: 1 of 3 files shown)


malavv:
Nice PR, lots of small additions. I didn't notice the lack of sound, and closing any player panels is also a good point.
Apart from that the Open, and Close implemented the `d2player.Panel` interface, and now its only partially fulfilled but I don't think that's a massive issue.

aziule:
> Apart from that the Open, and Close implemented the d2player.Panel interface, and now its only partially fulfilled but I don't think that's a massive issue.

@malavv you are 100% right, but I didn't see the `Panel` interface used anywhere. We could inject it instead of the plain implementation of `Inventory`, `HeroStats` or `EscapeMenu` in the `GameControls` struct.

PR #379, opened 2020-06-21
Map engine re-do

essial:
This massive PR completely overhauls the map engine. Previously maps were made of multiple 'regions'. Now a map is just that, a map. What was called 'regions' before, are now called 'stamps', which can be stamped onto the map at any position.

There's two outstanding issues that need to be fixed that are known in this PR:
- The mouse over debug info is not available in the map editor test.
- The start location isn't always at a good spot.

I'll create issues for these.


gravestench:
should all of the getter funcs for `MapEngine` in `d2core/d2map/d2mapengine/engine.go` be prefixed with `Get` ? like `GetTiles`

PR #390, opened 2020-06-22
fix for #358 from recent PRs

Haashi:
Recent PR updating maprenderer and adding escape menu erased #358 feature.


essial:
Sorry it was a bad merge, I erased my feature too..

PR #414, opened 2020-06-23
Added SuperUniques data dictionary

gurkankaymak:


essial:
I didn't see the part that actually loads the files, double-check that's happening. If not make a second PR for that. Thanks!

gurkankaymak:
you're right, forgot to call the load method from the main, second pr on the way

Issue #413, opened 2020-06-23
Bottom panel interactivity


nicholas-eden:
There's lots of things that can be done for the bottom panel:

- Fill the health and mana with a red/blue circle based on a percent input
- Trigger respective panels (or just log) when the user clicks on the buttons

essial:
This auto-closed. Re-opening as there is much to do.

essial:
#436

Natureknight:
Can we close this one?

essial:
We can, it seems to work well now!

PR #418, opened 2020-06-23
Handles Click and Hovers on most game control elements (#413)

malavv:
Creates a new type of "Actionable" which includes most of the bottom
bar elements. Then handles both "Click" and "Hover" into a separate
handler function.

At the moment, hover does nothing but routes correctly. Click logs
the corresponding actionable.

Known issues:
  - Not capturing button press on click (meaning player will move).
  - Basic touch detection, could be a more fancy QuadTree like struct.
  - Since the visual frames used are not just the actionable, the actual
    x, y coordinates of the actionable need to be entered statically. Changes
    in resolution would probably affect this.

### DIFF
--- a/d2game/d2player/game_controls.go
+++ b/d2game/d2player/game_controls.go
@@ -39,28 +39,28 @@ type GameControls struct {
 	FreeCam       bool
 
 	// UI
-	globeSprite     *d2ui.Sprite
-	mainPanel       *d2ui.Sprite
-	menuButton      *d2ui.Sprite
-	skillIcon       *d2ui.Sprite
-	zoneChangeText  *d2ui.Label
-	isZoneTextShown bool
-	actionables     []ActionableRegion
+	globeSprite       *d2ui.Sprite
+	mainPanel         *d2ui.Sprite
+	menuButton        *d2ui.Sprite
+	skillIcon         *d2ui.Sprite
+	zoneChangeText    *d2ui.Label
+	isZoneTextShown   bool
+	actionableRegions []ActionableRegion
 }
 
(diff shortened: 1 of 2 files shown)


essial:
Tag the issue this is for so it can be tracked. Thanks for the PR!

d2game/d2player/game_controls.go
```diff
@@ -64,6 +81,16 @@ func NewGameControls(hero *d2mapentity.Player, mapEngine *d2mapengine.MapEngine,
 		heroStats:      NewHeroStats(),
 		escapeMenu:     NewEscapeMenu(),
 		zoneChangeText: &label,
+		actionables: []ActionableRegion {
```
essial, line 84: Please change to `actionableRegions`

d2game/d2player/game_controls.go
```diff
@@ -44,8 +45,24 @@ type GameControls struct {
 	skillIcon       *d2ui.Sprite
 	zoneChangeText  *d2ui.Label
 	isZoneTextShown bool
+	actionables     []ActionableRegion
 }
 
+type Actionable int
```
essial, line 51: Change to `ActionableType`

d2game/d2player/game_controls.go
```diff
@@ -44,8 +45,24 @@ type GameControls struct {
...
+type ActionableRegion struct { AType Actionable; Rect d2common.Rectangle }
```
essial, line 52: Change `AType` to `ActionableTypeId` or something.

essial:
Looks great, thanks!

PR #419, opened 2020-06-23
Feature/player equipment ui

Ripolak:
Add rendering of player items and representations of Inventory equipped slots. Currently does not get the player's actual items, we should do it together with the inventory items.

Also currently we don't load rings and amulets as they are in the misc.txt item, which we currently don't render items from. We should add it to issues.

Small screenshot attached :)

INVENTORY


d2common/d2enum/inventory_item_type.go
```diff
@@ -3,7 +3,7 @@ package d2enum
 type InventoryItemType int
 
 const (
-	InventoryItemTypeItem   InventoryItemType = 0 // Item
+	InventoryItemTypeItem   InventoryItemType = 0 // item
```
essial, line 6: Why?

Ripolak: I believe this was added after I did a `git pull` last time, I fixed it now.

essial:
Did you want to add your name to CONTRIBUTORS?

PR #424, opened 2020-06-23
Feature/load misc items

Ripolak:
Allow rendering icons of misc items from the misc.txt file for viewing in inventory.


gravestench:
Aren't these already loaded by item_common ?

Ripolak:
> Aren't these already loaded by item_common ?

Not for rendering on screen from what I've seen, but as logical structs. I might be missing something though, do correct me if I'm wrong.

gravestench:
I made a stupid assumption instead of reading into your commit, apologies.

d2core/d2inventory/inventory_item_misc.go
```diff
@@ -0,0 +1,64 @@
+package d2inventory
+
+import (
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2data/d2datadict"
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2enum"
+	"log"
+)
+
+type InventoryItemMisc struct {
```
essial, line 9: Could we make this more generic and simply inherit properties? InventoryItemArmor and Weapon both use these fields as well, so it'd make sense to make a generic version and just extend for those.

Ripolak: Could be, those 3 structs actually already implement an `InventoryItem` interface (d2core/inventory/inventory_item.go), how'd you suggest we make it more generic?

essial: Gotcha, I'll accept for now, we can clean it up later. Thanks again!

PR #426, opened 2020-06-24
Update go.mod

ewohltman:
Remove replace directive


essial:
Wow must have missed that.. Good catch!

PR #429, opened 2020-06-24
Update to Go 1.14

ewohltman:


ewohltman:
Linting in CI is now working as expected.

There is a field in the workflow yamls, `continue-on-error: true` that allows the the CI build to continue even if the linting step fails. In the future, I'd recommend we remove `continue-on-error: true` to fail builds again.

This CI build is also failing due to existing tests failing, but I'd like to think fixing those is out of scope for this PR 😄

PR #442, opened 2020-06-24
added a startup error screen shown when the mpq files are not found

### LINKED ISSUE #416: Add a startup error screen
Currently the game will crash out if the mpq files are not located. We need the game to start up and render a helpful message about not being about to find the MPQs, as well as point the player to the directory where the config file is. I see this as a simple black screen with the helpful text centered in it using the built-in ebiten font.

gurkankaymak:
resolves #416 

resolved by creating a separate update function to pass to ebiten when the mpq files are not found, feel free to request/suggest any changes, I would update the pr accordingly

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,12 +64,6 @@ var singleton struct {
 }
 
 func main() {
-	if len(GitBranch) == 0 {
-		GitBranch = "Local Build"
-	}
-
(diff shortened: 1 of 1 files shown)


essial:
Nice, thanks so much!

Issue #400, opened 2020-06-22
Add run button


essial:
Now that run is working (press R to toggle), we need to render the run button the UI, and allow the user to toggle it with their mouse, in addition to the R key on the keyboard. The toggle state should reflect the run state of the user.

Natureknight:
Hey, I can take this one if that's fine for you

essial:
Go for it! Let me know if you decide not to do this so I can un-assign. Good luck!

PR #443, opened 2020-06-24
render the run button the UI, and allow the user to toggle it with th…

Natureknight:
…eir mouse, in addition to the R key on the keyboard (#400)


essial:
Great work, thanks!

PR #446, opened 2020-06-25
Switch to fork of coverage tool

joboscribe:
The existing suggestion for the code coverage viewer depends on the now unavailable ms-vscode.go extension.


essial:
Thank you!

Issue #351, opened 2020-06-20
Improve the loading screen


essial:
The loading screen used to animate, but we had to take that out. I'd like to either re-implement the animated loading screen, or at least a progress bar or something to make things smoother and look nicer.

carrelld:
I'll try this

essial:
You got it!

PR #447, opened 2020-06-25
default done loading after OnLoad finishes

carrelld:
No story with this one, but followup to #351 .

By doing a default loadState.Done(), we can be certain that even an empty `OnLoad` handler func won't cause a screen to hang.


carrelld:
Didn't test yet

PR #450, opened 2020-06-25
Refactor escape

brendanporter:
I moved the escape menu from `d2player` to `d2gamescreen` because it was necessary to call `CreateMainMenu()` to exit the game.

Now clicking SAVE AND EXIT GAME kicks you back to the loading screen, then main menu.

New bug introduced is the unexpected player permanence. If you leave the game and re-enter, the previous player is still present where you left them. Do this a number of times, moving each player to a different location, to simulate a party with synchronized idle animations. See screenshot below. Essial reports entity handling is likely changing tomorrow, so I will revisit this once that is done, as I am not yet familiar with how that is handled.
Screen Shot 2020-06-25 at 12 47 00 AM

It looks like gofmt modified a few extra lines that I hadn't originally noticed.

### DIFF
--- a/d2game/d2gamescreen/escape_menu.go
+++ b/d2game/d2gamescreen/escape_menu.go
@@ -1,4 +1,4 @@
-package d2player
+package d2gamescreen
 
 import (
 	"log"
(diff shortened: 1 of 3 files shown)


carrelld:
I think I recall you can send a local packet to tell the server someone has disconnected. The game client should have a callback interface that `MovePlayer` uses to communicate the movements to all connected clients. You could modify this interface to remove the player and entity from the game.

PR #457, opened 2020-06-25
Reorganize run.sh

### LINKED ISSUE #455: `run.sh` does not install dependencies if go is already installed
**Describe the bug**

When running `run.sh`, if `go` is already installed, the step which installs all dependencies is skipped.

**To Reproduce**
1. Have `go` already installed
2. Do not have some other dependency installed (for example, alsa-libs).
3. Execute `./run.sh` to build
4. It fails, because dependency is missing

**Expected behavior**

Program should build properly.

**Desktop (please complete the following information):**
 - OS: Fedora (although it happens with any Linux)
 - Version 31

**Additional context**

The `run.sh` script should have dependency split into two steps. First should install go. Second should install all other dependencies.

Also, the script is called `run.sh`, which implies that it runs the program, which it doesn't. It should probably be called `build.sh`.

goodevilgenius:
Fixes #455 

This PR reorganizes the `run.sh` script to ensure that system dependencies are installed.

It also rewords a few awkward wordings in that script.

Finally, it renames the script `build.sh`, since `run.sh` doesn't actually run OpenDiablo, but rather, builds it.


essial:
Thanks!

Issue #394, opened 2020-06-22
Add the options selection menu


essial:
When clicking options in game, it should show the following list:

SOUND OPTIONS
VIDEO OPTIONS
AUTOMAP OPTIONS
CONFIGURE CONTROLS
PREVIOUS MENU

Only thing that needs to be implemented is the "previous menu" option, though you should be able to click on the other options and it should print out that it was clicked like the main escape menu does today.

aziule:
Hey, I can take this one if that's fine for you

essial:
Assigned, let me know if you change your mind so I can un-assign. Good luck!

PR #459, opened 2020-06-25
Enhanced escape menu with options - d2gui bug remaining

aziule:
Partially solves #394 - small bug remaining that I'm unable to fix.

# Done
* Rewrite the Escape menu using `d2gui` as the foundations instead of hardcoding each element's position or recalculating it. It is especially useful when it comes to labels with several options possible, as well as with translations later on - `d2gui` places everything at the right place.
* It is now easy to update labels and handle callbacks when a user changes an option
* Support mouse (hovering + clicks) and keyboard (up + down + enter + escape)
* Add a few things into `d2gui` to support my changes. Some of them might be hacky so I'm open to suggestions here.

# To do
* There's a bug with the pentagrams that I couldn't figure: I can't render them properly when opening menu layouts for the first time (see screenshots below). It is related to how `d2gui` works internally but I'm not familiar enough with 2D engines so I can leave it up to someone who knows what to do.

# Screenshots
## The menu
Screenshot 2020-06-25 at 21 51 56

Screenshot 2020-06-25 at 21 52 06

## Support for editable options
Screenshot 2020-06-25 at 21 52 13

Screenshot 2020-06-25 at 21 52 24

## With debug mode
Screenshot 2020-06-25 at 21 54 51

## The bug with pentagrams
Screenshot 2020-06-25 at 21 55 28


d2core/d2gui/label.go
```diff
@@ -39,3 +33,25 @@ func (l *Label) render(target d2render.Surface) error {
...
+func (l *Label) GetText() string {
+	return l.text
+}
+
+func (l *Label) SetText(text string) error {
+	return l.setText(text)
+}
+
+func (l *Label) setText(text string) error {
+	width, height := l.font.GetTextMetrics(text)
```
essial, line 46: This needs to simply return if they text has not changed. No reason to re-generate the graphics if the change is a no-op.

essial:
Great work! For the remaining issues, can you create a bug for that?

d2core/d2gui/label.go
aziule: good catch 👍

d2core/d2gui/widget.go
```diff
@@ -44,6 +49,23 @@ func (w *widgetBase) SetPosition(x, y int) {
...
 
+func (w *widgetBase) GetPosition() (int, int) {
+	return w.x, w.y
+}
+
+func (w *widgetBase) GetOffset() (int, int) {
+	return w.offsetX, w.offsetY
+}
+
+func (w *widgetBase) getOffset() (int, int) {
```
carrelld, line 60: Why is this duplicated as an unexported method?

essial: Good question. I'm looking into that last bug, I'll remove this.

aziule: At first I wanted to keep the interface methods unexported but you are right, it looks like an unnecessary duplication => I only kept the exported one.

malavv:
I think the last bug is for all intent and purposes difficult to reproduce since PR #559
Now that the general mouse region selects an element. An element is selected near the first frame.

Issue #412, opened 2020-06-23
Player animations for missile casting


nicholas-eden:
When the player right clicks the hero should stop moving and go through the cast animation.  When the animation is finished, the current missile should spawn.

carrelld:
I guess I'll try this one now

carrelld:
Looks like the missile code currently implemented in `game_controls.go` but `MovePlayer` is implemented in the `game_client.go` behind the server/client stuff. Where did you imagine this being implemented?

essial:
There is currently no mechanism that I know of to handle casting animations. This will have to be implemented. I would imagine we send a move player command with the same start and destination location (or a 'place player' command that puts it in a specific spot, and puts them into a standing state and clears their path queue). Then we need a network packet for casting missiles. When that packet comes in we'll need a way to tell the entity to start the casting animation (as well as creating the actual missiles).

carrelld:
Can it be a single packet? Like CastPacket{skillID, sourceX, sourceY, targetX, targetY} from this, we know exactly where the player is when casting and can position/move them to sourceX,Y. That way we can avoid two packets. 

Let me know if this makes no sense. I'm feeling a bit in the deep end with this project and good game networking practice.

essial:
I'm think there's possibly two destination types.. either a specific X/Y (subtile), OR an entity ID. For source I don't think there's a situation where it wouldn't be coming from some kind of entity. So I'd say have the packet have a source entity ID, and either a destination x/y or destination entity.

presiyan-ivanov:
I implemented this logic with #729 - the issue can be closed in my opinion.

gravestench:
@presiyan-ivanov good work, and thank you!

PR #470, opened 2020-06-26
412 missle animation

carrelld:
Attempt at addressing https://github.com/OpenDiablo2/OpenDiablo2/issues/412

This sort-of works? I'm not happy with the way entity states are being tracked. I think we can create some sort of `Action` abstraction that can handle all the things an entity can do. That way we can orchestrate how they interact and what the rules are for things like canceling animations, when to fire a missile, etc. I think for now, this is a good enough start. Please feel free to disagree. I recommend playing with it a bit to see all the behaviors.

Known issues. 
*The full animation is not played if casting while moving
* Facing directly down performs the entire animation, while facing any other direction clips it shorter
* The hero doesn't rotate to face the direction of the casting target
* The missile shoots at the start of the animation instead of at the end (or some arbitrary point during the animation... I'm not sure what the correct behavior is
* Missiles are persisted on the map after they travel a certain distance and removing them is not currently implemented

### DIFF
--- a/d2core/d2map/d2mapengine/engine.go
+++ b/d2core/d2map/d2mapengine/engine.go
@@ -209,7 +209,7 @@ func (m *MapEngine) RemoveEntity(entity d2mapentity.MapEntity) {
 	if entity == nil {
 		return
 	}
-	panic("Removing entities is not currently implemented")
+	//panic("Removing entities is not currently implemented")
 	//m.entities.Remove(entity)
 }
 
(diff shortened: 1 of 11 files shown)


nicholas-eden:
I think this is a good start.  I think we need a system of chainable actions, ex: walk to npc then interact, animate then cast, run to enemy npc then melee.

I'm a bit fuzzy on how D2 worked exactly, but I'm pretty sure it behaved like that.

PR #477, opened 2020-06-28
Load monpreset and use that to find monstat entries for NPC's

Ziemas:
Then get NPC names by using the translation lookup string from monstat


essial:
Nice work!

PR #497, opened 2020-06-30
D2networking resolve lint issues

danhale-git:
This PR adds simple error reporting and creates a lot of mess. I will be doing one more pass to tidy things up a bit before I move on.


d2networking/d2client/d2localclient/local_client_connection.go
```diff
@@ -47,7 +49,7 @@ func Create(openNetworkServer bool) *LocalClientConnection {
 }
 
 // Open creates a new GameServer, runs the server and connects this client to it.
-func (l *LocalClientConnection) Open(connectionString string, saveFilePath string) error {
+func (l *LocalClientConnection) Open(_ string, saveFilePath string) error {
```
Natureknight, line 52: Can we refactor this a bit in order to remove this parameter as it seems we don't really need it?

danhale-git:
Found an issue with this PR, brb.

d2networking/d2client/d2localclient/local_client_connection.go
danhale-git: The parameter is necessary to satisfy an interface, hence why I've left it with the blank identifier.

danhale-git:
It looks like the remote client's packets are causing a gzip error.

```remote_client_connection.go:157: RemoteClientConnection: error copying bytes from AddPlayer packet: gzip: invalid header
remote_client_connection.go:157: RemoteClientConnection: error copying bytes from AddPlayer packet: gzip: invalid header
remote_client_connection.go:157: RemoteClientConnection: error copying bytes from Ping packet: gzip: invalid header
remote_client_connection.go:157: RemoteClientConnection: error copying bytes from Ping packet: gzip: invalid header
remote_client_connection.go:157: RemoteClientConnection: error copying bytes from Ping packet: gzip: invalid header

I think this the case in master too, we just hadn't noticed due to lack of error handling. It would explain some odd behaviour I've been seeing when testing multiplayer.

@essial  The game runs now, in single and multiplayer. Seems like the MP client can't move but I think that's the case in master too, potentially related to these invalid header errors. On that note I'll reopen the pull to get these changes out the way. Feel free to close it out if you'd rather not merge until those errors are fixed. The next PR will be refactoring/tidying.

File: d2networking/d2client/d2localclient/local_client_connection.go (hunk 1 above)
Comment (danhale-git) [reply d2networking/d2client/d2localclient/local_client_connection.go:52]: Possibly something that warrants a comment, I initially removed the parameter myself.

**Comment by essial** [comment]
MP client should be able to move, the only difference is that it's using an external client.

File: d2networking/d2netpacket/d2netpackettype/message_type.go [hunk 2]
```diff
@@ -26,3 +28,30 @@ const (
 	ServerClosed                                         // Sent by the local host when it has closed the server
 	CastSkill                                            // Sent by client or server, indicates entity casting skill
 )
+
+func (n NetPacketType) String() string {
```
essial, line 32: Can we make this a map?

danhale-git: Absolutely, I will make that change.

I believe implementing [Stringer ](https://golang.org/pkg/fmt/#Stringer) is idiomatic, hence why Go doesn't ask us for a 
comment on that method. Out of curiosity, what is the benefit of a map?

essial: If we have 50 packet types, I'd rather do a hash match then a ton of CPU branches. It's not a problem at the moment, but down the road it can be.

essial: If anyone disagrees feel free to speak up!

danhale-git: Done. I was mistaken, using a map doesn't mean not implementing Stringer, it's still implemented but references the map in place of the switch.

danhale-git:
@essial This can be merged at your discretion. I moved the map into the struct for now.

I had to roll back resolution of the remaining warnings. I will get them done again with the next bit of work. I'm seeing some strange behaviour with linting and Goland (a persistent, unwarranted type check error) so I want to start fresh before I continue.

d2networking/d2client/d2localclient/local_client_connection.go
```diff
@@ -57,7 +59,10 @@ func (l *LocalClientConnection) Open(connectionString string, saveFilePath strin
 
 // Close disconnects from the server and destroys it.
 func (l *LocalClientConnection) Close() error {
-	l.SendPacketToServer(d2netpacket.CreateServerClosedPacket())
+	err := l.SendPacketToServer(d2netpacket.CreateServerClosedPacket())
+	if err != nil {
+		log.Printf("LocalClientConnection: error sending ServerClosedPacket to server: %s", err)
```
carrelld, line 64: What's the point of swallowing the error? I'd prefer to see the lint warnings over swallowing it, unless we truly do just want to ignore it.

danhale-git: I was under the impression we were trying to remove all lint warnings. I will be doing more work on d2networking during which I will have a proper look at the error handling.

danhale-git: On second thoughts, I could have just handled the errors as opposed to using the blank identifier. I will fix that.

PR #501, opened 2020-06-30
Fixes #496 and improves memory consumption on render of game_controls

### LINKED ISSUE #496: Rendering sections of surfaces without allocating new images.
Globes work by rendering part of their sprite. Since we only have Render() at the moment, the only way to Render part of a sprite is to create a smaller image, render the full image in the smaller surface, print the smaller surface, and free the temporary smaller surface.

I believe it would be logical to have the ability to ask to render only a section of a surface. On the globes, it would remove 2 new surface on each render. But I suspect this would be useful elsewhere.

**Describe alternatives you've considered**
I've tried for 2 days to get "masking" to work. Where, using another image, I could select by drawing pixels with different opacity, what part of the image to render. Seems like it should work, but I kept getting stuck with surfaces that flatten the image and removes the alpha during the blending. And therefore big black squares instead of transparencies. But even this method is somewhat of a hack versus just drawing what you need.

malavv:
The main contributions are in `ebiten_surface.go` and `surface.go`. I added methods in Sprite and Animation to access the new surface functionality.

I tested for allocations and they seem normal, removed all old code to circumvent this issue, and fixed comment and spacing linting issues in Sprite and Animation.

### DIFF
--- a/d2common/d2interface/surface.go
+++ b/d2common/d2interface/surface.go
@@ -23,6 +23,8 @@ type Surface interface {
 	PushTranslation(x, y int)
 	PushBrightness(brightness float64)
 	Render(surface Surface) error
+	// Renders a section of the surface enclosed by bounds
+	RenderSection(surface Surface, bound image.Rectangle) error
 	ReplacePixels(pixels []byte) error
 	Screenshot() *image.RGBA
 }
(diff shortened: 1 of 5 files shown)


essial:
Looks great, thanks!

PR #510, opened 2020-07-01
Fixes #509, Creates a Allocation Rate Debug Metric.

### LINKED ISSUE #509: Print the Total Allocation Rate/s in Debug Overlay
As we are looking to investigate memory usage, looking at "how quick alloc increases" is difficult.

I would want the rate of increase to be one of the measure. For the moment, in MB/s. This would have to be a running mean to be meaningful.

**Describe alternatives you've considered**
Looking at it and do the math in my head.

**Additional context**
Probably we can print this value in the Debug/fps screen overlay.

malavv:
In order to be stable, it needs to collect many samples. Hence, I use a Ring to always know what was the Total Allocation 100 frames before.

### DIFF
--- a/d2app/app.go
+++ b/d2app/app.go
@@ -3,6 +3,7 @@ package d2app
 
 import (
 	"bytes"
+	"container/ring"
 	"errors"
 	"fmt"
 	"image"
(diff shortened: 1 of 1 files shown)


malavv:
Here is the preview.
image

Issue #467, opened 2020-06-26
Remove pathfinding logic and enhance movements


essial:
Today we are using A* pathfinding, but it's creating a ton of problems for absolutely no reason. There are 16 directional animations, but there can only be 8 connecting nodes to a subtile, so this solution is not viable. 90% of the time pathfinding isn't necessary when playing the game, and there seems to be only a few instances where it's even used. We can leave the existing A* functions in for later use when we get around to it, but the actual pathfinding needs to go.

For now we need to totally remove pathfinding and the walk mesh. The From/To player movement commands should stay, but the player should move directly towards that point until it hits a non-walkable subtile, at which point they should just stop there.

That should allow way more natural and smooth movement in the game.

Issue #452, opened 2020-06-25
Memory keeps growing when re-starting a game


essial:
If you keep leaving and starting a game, memory usage keeps climbing. I've attached a heap dump after a few restarts.
[heap.pdf](https://github.com/OpenDiablo2/OpenDiablo2/files/4831905/heap.pdf)

essial:
Reset map does this, which I feel is leaking tons of memory:
`	m.entities = make([]d2mapentity.MapEntity, 0)
	m.levelType = d2datadict.LevelTypes[levelType]
	m.size = d2common.Size{Width: width, Height: height}
	m.tiles = make([]d2ds1.TileRecord, width*height)
	m.dt1TileData = make([]d2dt1.Tile, 0)
	m.walkMesh = make([]d2common.PathTile, width*height*25)
	m.dt1Files = make([]string, 0)

	for idx := range m.levelType.Files {
		m.addDT1(m.levelType.Files[idx])
	}`

presiyan-ivanov:
I can take this issue

essial:
Assigned, sorry I apparently missed your message earlier!

DanMaute:
Do you think the go runtime is actually leaking memory or not releasing it to the OS? It may be that the go runtime is garbage collecting the memory but is just not releasing the memory to the OS as fast as expected. I haven't studied the heap dump too closely or looked at the offending code, but the symptoms as described remind me of an issue I have seen in the past.

Not a good practice but you could call debug.FreeOSMemory to force garbage collection and subsequent release of the unused memory to the OS. Doing so may conceal underlying memory management issues, but it could be useful for debug purposes if the issue is obstructing development.

These describe what I'm referring to:

[https://stackoverflow.com/questions/42345060/freeosmemory-in-production](url)

[https://stackoverflow.com/questions/37382600/cannot-free-memory-once-occupied-by-bytes-buffer/37383604#37383604](url)

[https://github.com/golang/go/issues/22439](url)

gravestench:
@presiyan-ivanov what is the status of this issue?

PR #513, opened 2020-07-01
Reduce memory growth when re-starting a game (#452)

presiyan-ivanov:
- Dispose of images when leaving a scene to prevent ebiten copying
pixels of the images

- Remove walkmesh as it will be redundant(see #467) and consumes a lot
of memory

- Use only index in for loops to prevent value getting copied

- Save and exit now stops the local server


presiyan-ivanov:
Some of the scene transitions seem to be broken after the changes(char select -> main menu and char select -> create hero). I will fix those and re-open

PR #516, opened 2020-07-01
Lint fixes for the d2game/d2gamescreen

gurkankaymak:
There are many lint issues in the d2gamescreen package, it'll take some time, so I'm commiting file by file
currently the following files are fixed
`escape_menu.go`
`credits.go`


gravestench:
Looks good to me

PR #519, opened 2020-07-01
Reduce memory growth when re-starting a game (#452)

presiyan-ivanov:
- Dispose of images when leaving a scene to prevent ebiten copying
pixels of the images

- Widgets now automatically get disposed on unloading a scene

- Remove walkmesh as it will be redundant(see #467) and consumes a lot
of memory

- Use only index in for loops to prevent value getting copied

- Save and exit now stops the local server


d2common/d2data/d2datadict/objects.go
```diff
@@ -344,9 +344,10 @@ var Objects map[int]*ObjectRecord
 func LoadObjects(file []byte) {
 	Objects = make(map[int]*ObjectRecord)
 	data := strings.Split(string(file), "\r\n")[1:]
+	for idx := range data {
+		line := data[idx]
```
malavv, line 348: Same comment for all similar modifications, but are you sure this is changing the compiled executable? Seems exactly like the kind of optimization I expect the compiler to do.

d2common/cache.go
```diff
@@ -40,6 +40,18 @@ func (c *Cache) GetBudget() int {
 	return c.budget
 }
 
+func (c *Cache) GetKeys() []string {
+	keys := make([]string, len(c.lookup))
+
+	i := 0
+	for k := range c.lookup {
```
gravestench, line 47: why not 
```golang
for i := 0; i < len(c.lookup); i++ { ... }
```

d2common/d2data/d2datadict/objects.go
essial: > Same comment for all similar modifications, but are you sure this is changing the compiled executable? Seems exactly like the kind of optimization I expect the compiler to do.

We've seen it actually create copies when doing the two-value for range loops. So I prefer we just use indexes and not worry about it.

d2core/d2asset/animation.go
```diff
@@ -454,11 +459,32 @@ func (a *Animation) ResetPlayedCount() {
 	a.playedCount = 0
 }
 
+// Dispose removes the animation to ensure we do not hold the underlying images
```
gravestench, line 462: lint error

d2core/d2asset/animation.go
```diff
@@ -454,11 +459,32 @@ func (a *Animation) ResetPlayedCount() {
...
+
+		for idx := range dir.frames {
+			dir.frames[idx].image.Dispose()
+		}
+	}
+
+	UnloadAnimation(a.key)
+}
+
+func (a *Animation) SetKey(key string) {
```
gravestench, line 475: lint error

d2core/d2asset/animation.go
```diff
@@ -454,11 +459,32 @@ func (a *Animation) ResetPlayedCount() {
...
+	a.key = key
+}
+
+func (a *Animation) GetKey() string{
```
gravestench, line 479: lint error

all of these exported functions should have comments

d2common/d2data/d2datadict/objects.go
malavv: The spec seem to say it shouldn't 
https://golang.org/ref/spec "For statements with range clause"
```
Range expression                          1st value          2nd value

array or slice  a  [n]E, *[n]E, or []E    index    i  int    a[i]       E
string          s  string type            index    i  int    see below  rune
map             m  map[K]V                key      k  K      m[k]       V
channel         c  chan E, <-chan E       element  e  E
```
It says the second value is literally generated using the `data[idx]` pattern

essial: https://play.golang.org/p/5LG71jC5tTj

presiyan-ivanov: Yep, I've tested it, you can check it here: https://play.golang.org/p/2GMYyC3Ai7b

essial: I was wrong. This is incorrect. when you say `line := data[idx]` you have created a copy. The correct syntax would be `line := &data[idx]`

d2common/cache.go
```diff
@@ -40,6 +40,18 @@ func (c *Cache) GetBudget() int {
 	return c.budget
 }
 
+func (c *Cache) GetKeys() []string {
+	keys := make([]string, len(c.lookup))
```
malavv, line 44: I think there might be a race conditions here, as even reading, the array could be modified while iterating it. I'm only saying that because I thing the class is supposed to be semi threadsafe

d2core/d2asset/animation_manager.go
```diff
@@ -20,10 +20,27 @@ func createAnimationManager() *animationManager {
 	return &animationManager{d2common.CreateCache(animationBudget)}
 }
 
+// Dispose of all the animations to prevent ebiten storing the images
+func (am *animationManager) ClearCache() {
+	keys := am.cache.GetKeys()
+	for idx := range keys {
+		key := keys[idx]
```
gravestench, line 27: this is making a local copy of `keys[idx]`, i think it should be `key := &keys[idx]`

d2core/d2asset/composite.go
```diff
@@ -37,7 +37,8 @@ func (c *Composite) Advance(elapsed float64) error {
 	c.mode.playedCount += c.mode.frameIndex / c.mode.frameCount
 	c.mode.frameIndex %= c.mode.frameCount
 
-	for _, layer := range c.mode.layers {
+	for idx := range c.mode.layers {
+		layer := c.mode.layers[idx]
```
gravestench, line 41: I think this is making a local copy of `c.mode.layers[id]`

d2core/d2asset/composite.go
```diff
@@ -54,7 +55,8 @@ func (c *Composite) Render(target d2interface.Surface) error {
 		return nil
 	}
 
-	for _, layerIndex := range c.mode.drawOrder[c.mode.frameIndex] {
+	for idx := range c.mode.drawOrder[c.mode.frameIndex] {
+		layerIndex := c.mode.drawOrder[c.mode.frameIndex][idx]
```
gravestench, line 59: i think this is making a local copy of `c.mode.drawOrder[c.mode.frameIndex][idx]`

d2core/d2asset/composite.go
```diff
@@ -172,7 +174,8 @@ func (c *Composite) createMode(animationMode, weaponClass string, direction int)
 		mode.drawOrder[frame] = cof.Priority[mode.cofDirection][frame]
 	}
 
-	for _, cofLayer := range cof.CofLayers {
+	for idx := range cof.CofLayers {
+		cofLayer := cof.CofLayers[idx]
```
gravestench, line 178: i think this is making a local copy of `cof.CofLayers[idx]`

d2core/d2map/d2mapgen/act1_overworld.go
```diff
@@ -64,8 +63,6 @@ func GenerateAct1Overworld(mapEngine *d2mapengine.MapEngine) {
 		// North Exit
 		mapEngine.PlaceStamp(townStamp, mapWidth-townSize.Width, mapHeight-townSize.Height)
 	}
-
-	mapEngine.RegenerateWalkPaths()
```
malavv, line 68: Could removing this cause issue with walking in the game?

d2core/d2map/d2maprenderer/tile_cache.go
```diff
@@ -40,22 +40,22 @@ func (mr *MapRenderer) generateTileCache() {
...
 		log.Printf("Could not locate tile Style:%d, Seq: %d, Type: %d\n", tile.Style, tile.Sequence, 0)
-		tileData = append(tileData, &d2dt1.Tile{})
+		tileData = append(tileData, d2dt1.Tile{})
 		tileData[0].Width = 10
 		tileData[0].Height = 10
 	} else {
 		if !tileOptions[0].MaterialFlags.Lava {
 			tileIndex = mr.getRandomTile(tileOptions, tileX, tileY, mr.mapEngine.Seed())
-			tileData = append(tileData, &tileOptions[tileIndex])
+			tileData = append(tileData, tileOptions[tileIndex])
```
malavv, line 54: Maybe I'm over-suspicious because of the assignment thing, but doesn't this copy the like option at index "tileIndex"? Since it's now a slice of struct instead of *struct?

d2core/d2map/d2mapgen/act1_overworld.go
presiyan-ivanov: Walking is not working after removing walkmesh, I was thinking of implementing it in a separate PR but I think @essial  will work on that and I will sync with his changes after that.

d2core/d2asset/animation_manager.go
```diff
@@ -20,10 +20,27 @@ func createAnimationManager() *animationManager {
...
+	}
+
+	am.cache.Clear()
+}
+
 func (am *animationManager) loadAnimation(animationPath, palettePath string, transparency int) (*Animation, error) {
 	cachePath := fmt.Sprintf("%s;%s;%d", animationPath, palettePath, transparency)
 	if animation, found := am.cache.Retrieve(cachePath); found {
-		return animation.(*Animation).Clone(), nil
+			return animation.(*Animation), nil
```
Ziemas, line 43: Will this not make all animations of the same type point at the same object?

Keep in mind that there can be multiple instances of the same animation loaded but with different properties (speed, direction, etc).

essial: Wouldn't returning the dereference cause it to copy the object? The gotcha here would be if the animation object had pointers -- if it does, then this would retain the shared data erroneously.

d2common/cache.go
presiyan-ivanov: Actually it cannot be done with the approach you proposed, since I need a for-each to iterate over the map keys of c.lookup in order to store them at the i-th pos of the keys array.

gravestench: that's a good why not

PR #521, opened 2020-07-02
Proposal for Vector interface using `big.Float` components (with test implementation)

gravestench:
_I'll start by saying that this Vector2 interface is not complete, as I'm not sure how exactly to do some things. In particular, I'm not sure how we should be marshaling the struct with the `big.Float`s, as well as how to perform things like sin/cos on the `big.Float` objects, but I'd like for us to talk about it._

@essial has been wanting to clean up the coordinate system(s) that we currently have in place, and yesterday I spent most of my day making a `World/Tile/Subtile` coordinate system. However, @malavv suggested we should be using a vector coordinate system like other game engines do (_a sound, fucking valid point_), and I agree.

Anyways, what I'm submitting here doesn't actually account for translating from screen / orthogonal / tile / subtile, as that really belongs somewhere else.

**The real point here is that this implementation of a 2-dimensional vector uses an infinite precision floating point number representation**. As we've all heard by this point, the loss of information due to rounding errors from floating point numbers makes everybody uneasy. Hopefully, this vector implementation can account for that.

Along with this commit, I'm also submitting some **additional interfaces**.

- Serializer -- something that can `Marshal` and `Unmarshal` data
- WorldEntity -- something with a `Position()` method that yields a `Vector2` (inherits Seriailizer)

The justification for these two interfaces is that all map entities which game-clients need to know about will need the ability to marshal / unmarshal data, and that should be described one way to make our lives easy.

I've tried to be fairly thorough, this should have only a few linting errors and should not interfere with builds.

### DIFF
--- a/d2common/d2data/d2datadict/objects.go
+++ b/d2common/d2data/d2datadict/objects.go
@@ -9,6 +9,7 @@ import (
 
 // An ObjectRecord represents the settings for one type of object from objects.txt
 type ObjectRecord struct {
+	Index         int    // Line number in file, this is the actual index used for objects
 	FrameCount    [8]int // how many frames does this mode have, 0 = skip
 	FrameDelta    [8]int // what rate is the animation played at (256 = 100% speed)
 	LightDiameter [8]int
@@ -21,7 +22,7 @@ type ObjectRecord struct {
 	Description string
 	Token       string // refers to what graphics this object uses
 
-	Id              int //nolint:golint it's ok that it's called Id
+	Id              int //nolint:golint it's ok that it's called Id, unused indexed by line number instead
 	SpawnMax        int // unused?
 	TrapProbability int // unused
 
@@ -345,6 +346,8 @@ func LoadObjects(file []byte) {
 	Objects = make(map[int]*ObjectRecord)
 	data := strings.Split(string(file), "\r\n")[1:]
 
+	lineNumber := 0
+
 	for _, line := range data {
 		if line == "" {
 			continue
@@ -357,7 +360,9 @@ func LoadObjects(file []byte) {
 		}
 
 		rec := createObjectRecord(props)
-		Objects[rec.Id] = &rec
+		rec.Index = lineNumber
+		Objects[lineNumber] = &rec
+		lineNumber++
 	}
 
 	log.Printf("Loaded %d objects", len(Objects))


danhale-git:
I like the look of this.

I would lean towards calling it either 'Float2' or 'Vector'. Assuming we won't need a 3-value struct, Vector would suffice since a vector by definition has multiple values (I think).

gravestench:
I'm okay with calling it `Vector`, and for the `WorldEntity` interface the method could just be called `Vector` for retrieving it.

danhale-git:
I think `entity.Position()` reads well.

It is also worth noting that Go has Marshaler and Unmarshaler interfaces, which I believe the JSON package will look for when Marshal/Unmarshal is called.
https://golang.org/pkg/encoding/json/#Unmarshaler
https://golang.org/pkg/encoding/json/#Marshaler

It seems like Go can unmarshal and marshal big floats out of the box though: https://play.golang.org/p/7XKn2hhXRD

gravestench:
I didn't know if we would be sticking with json, but that's way easy.

d2common/d2math/vector2.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+// Angle computes the angle in radians with respect
+// to the positive x-axis
+func (v *Vector2) Angle() *big.Float {
+	// HACK we should find a way to do this purely
+	// with big.Float
+	floatX, _ := v.X().Float64()
+	floatY, _ := v.Y().Float64()
+	floatAngle := math.Atan2(floatX, floatY)
```
malavv, line 151: I believe the correct call to math.Atan2 is `func Atan2(y, x float64) float64`

d2common/d2math/vector2.go
```diff
@@ -0,0 +1,367 @@
...
+	v.x.Sub(v.x, s)
+	v.y.Sub(v.y, s)
+
+	return v
+}
+
+// Divide this Vector2 by the given Vector2
+func (v *Vector2) Divide(src d2interface.Vector2) d2interface.Vector2 {
+	v.x.Quo(v.x, src.X())
+	v.y.Sub(v.y, src.Y())
```
malavv, line 200: Quo?

d2common/d2math/vector2.go
```diff
@@ -0,0 +1,367 @@
...
+func (v *Vector2) DistanceSq(src d2interface.Vector2) *big.Float {
+	delta := src.Clone().Subtract(v)
+	deltaSq := delta.Multiply(delta)
+
+	return big.NewFloat(zero).Add(deltaSq.X(), deltaSq.Y())
+}
+
+// Length returns the length of this Vector2
+func (v *Vector2) Length() *big.Float {
+	xsq, ysq := v.LengthSq()
```
malavv: (Maybe I don't get what this function does.) It probably doesn't matter, but I think this is called a `norm` or Euclidian distance, and not a length. On the other hand, people will 100% understand the term length.
If this function is supposed to return the norm of a vector, it's missing a square root.

d2common/d2math/vector2.go
```diff
@@ -0,0 +1,367 @@
...
+	one := big.NewFloat(1.0)
+
+	if length.Cmp(one) > 0 {
+		length.Quo(one, length.Sqrt(length))
+
+		v.x.Mul(v.x, length)
+		v.y.Mul(v.y, length)
+	}
+
+	return v
```
malavv: I think this might have been made clearer if we had a Length/Norm you don't need to sqrt. Also, I think Normalize should also normalize vector with a Norm of less than one. Making sure every input vector has a Magnitude or norm of 1 afterward.

d2common/d2math/vector2.go
gravestench: `Quo` is right, but `Sub` is definitely not! thanks for pointing that out

d2common/d2math/vector2.go
```diff
@@ -0,0 +1,367 @@
...
+	return c.X().Add(c.X(), c.Y())
+}
+
+// Cross Calculate the cross product of this Vector2 and the given Vector2.
+func (v *Vector2) Cross(src d2interface.Vector2) *big.Float {
+	c := v.Clone()
+	c.X().Mul(c.X(), src.X())
+	c.Y().Mul(c.Y(), src.Y())
+
+	return c.X().Sub(c.X(), c.Y())
```
malavv: I'm not super familiar with cross product in 2D space, but from other implementations I have seen do something like `(lhs.X*rhs.Y) - (lhs.Y*rhs.X)` (note (X * Y) and not (X * X)) but again, I'm not even sure what it means in 2D space.

d2common/d2math/vector2.go
```diff
@@ -0,0 +1,367 @@
...
+// Reflect this Vector2 off a line defined by a normal.
+func (v *Vector2) Reflect(normal d2interface.Vector2) d2interface.Vector2 {
+	clone := v.Clone()
+	clone.Normalize()
+
+	two := big.NewFloat(2.0) // there's some matrix algebra magic here
+	dot := v.Clone().Dot(normal)
+	normal.Scale(two.Mul(two, dot))
+
+	return v.Subtract(normal)
```
malavv: Could it be (normal - v) instead of v - normal ?
From Godot::Vector2 `return 2.0 * p_normal * this->dot(p_normal) - *this;`

PR #537, opened 2020-07-04
New package d2object, first object initFun

Ziemas:
Moves objects out to their own package and implements the very first init function, torches/braziers now gets their animation mode set at creation.


gravestench:
good work, ziemas

PR #546, opened 2020-07-05
App components

gravestench:
The way it works is that component packages have a provider function that creates an component, but does not immediately initialize it. Instead, the App is now responsible for initializing the app components, and this is done automatically within `app.Run`.


d2core/d2asset/asset_manager.go
```diff
@@ -1,23 +1,228 @@
...
+
+	case d2interface.ArchivedFontManager:
+		if err := manager.Bind(am); err != nil {
+			return err
+		}
+
+		am.fontManager = manager.(d2interface.ArchivedFontManager)
+	}
+
+	return nil
```
essial, line 169: Panic on the default case

gravestench: this gets bubbled up to `app.Initialize`, which gets bubbled up to `app.Run`, which throws a fatal error

d2common/d2interface/app.go
```diff
@@ -0,0 +1,24 @@
...
+	// The AppComponents
+	Input() (InputManager, error)
+	Audio() (AudioProvider, error)
+	Renderer() (Renderer, error)
+	Terminal() (Terminal, error)
+	Asset() (AssetManager, error)
+}
+
+// AppComponent is part of the app
+// components are first bound, then initialized.
```
mewmew, line 19: Either add blank line, or period and capital letter, as `go doc github.com/OpenDiablo2/OpenDiablo2/d2common/d2interface.AppComponent` would otherwise render the docs as follows:

*AppComponent is part of the app components are first bound, then initialized.*

gravestench: good catch, will fix.

gravestench: should be fixed :)

d2core/d2asset/asset_manager.go
gravestench: nevermind, I see what you're talking about. adding it.

gravestench:
Would appreciate feedback on this as 

@mewmew like you said, there are different  ways we could do asset management (like using archives and filesystem directories, or fetching over the net).

Also, I know we have a ton of interfaces at the moment, and **I'm all for removing them**, but at the time i had started introducing them it seemed necessary in order to decouple everything (was getting lots of import cycles). I think the next step after this would be to move all the interface declarations to where they are used.

mewmew:
> Would appreciate feedback on this.

Just a quick heads up. I'll be busy for the next three weeks, so it's probably better to ping someone who can move this along.

I understand if interfaces have to be used in some places to fix cyclic imports. So long as we don't start using interfaces, just to use interfaces. *"premature abstractions is the root of all evil"*, and all that.

malavv:
For what it's worth. I've looked at the modifications and they seem reasonable. I don't have the greatest familiarity with this code, but nothing jumped at me as wrong or wrong headed. So to me, it looks good.

There is a lot of redundancies in the code, and I've seen you mention hard ordering requirements. But I think going deeper in abstraction might not give a clear reward versus just having comments and never touching the ordering in app.go.

PR #547, opened 2020-07-05
Init torches with random start mode

Ziemas:
Just saw this was a thing and added it for fun.


Ziemas:
I just put this in my other object stuff PR instead.

PR #568, opened 2020-07-09
bugfix: strings from txt file are all lowercase

gravestench:


gravestench:
@Intyre will include this in his upcoming PR

PR #579, opened 2020-07-12
Map entity rework - vectors in mapEntity

danhale-git:
This PR implements `d2vector.Position` and `d2vector.Vector` in `mapEntity`. Apologies for the size, I tried to keep it small.

Most of the maths used in `mapEntity.Step()` has been completely replaced with the new `Vector` methods. I was careful to ensure parity before migrating from the old systems, so everything should still work as it did before.

There is more work to do here, hopefully this is a solid start.

Next up (probably)

* Profiling, testing and optimising `Position` and `mapEntity`.
* Review and consolidate the other types in `d2mapentity`.
* _something about map generation_

### DIFF
--- a/d2common/d2math/d2vector/position.go
+++ b/d2common/d2math/d2vector/position.go
@@ -99,11 +99,10 @@ func (p *Position) Offset() *Vector {
 	return &v
 }
 
-// TODO: rename this to RenderOffset
-// RenderOffset is SubTileOffset() + 1. This places the vector in at the bottom vertex of this sub tile, which is a
-// diamond in the isometric space.
-// SubTile indices increase to the lower right diagonal ('down') and to the lower left diagonal ('left').
-func (p *Position) SubCell() *Vector {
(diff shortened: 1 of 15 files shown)


danhale-git:
@gravestench Thanks for the feedback, appreciate it :) 

Well spotted on the math typo (I assume you meant the incorrect spelling of _mathematical_).

`Position `exists to bridge the gap between our normal sensible vector code and whatever madness Diablo 2 does with position/movement/rotation. I have considered a `Rotation `type also, in the same vein, but not sure it's needed. The plan is, once ~everything uses `Position`, we can have a clear picture of all that code and start refactoring accordingly. This might result in us removing `Position` altogether and having the systems which used it manage their own idiosyncrasies. So, consider Position as a refactoring tool in it's current state. Something to carry the burden of Diablo 2 coordinate systems.

That said, I am as happy moving those constants out as I am waiting to see what the future holds. Be that duplicate variables cropping up or those ones moving into other places naturally as things are consolidated. I welcome any advice on where to put them, I probably need to dig into the related code a bit more before I can make an informed decision myself.

PR #581, opened 2020-07-13
adding ranged number type, for use in stats

gravestench:
unit tests included

### DIFF
--- a/d2common/d2math/ranged_number.go
+++ b/d2common/d2math/ranged_number.go
@@ -1,0 +1,101 @@
+package d2math
+
+import "fmt"
+
+// RangedNumber is a number with a min and max range
+type RangedNumber struct {
+	min int
+	max int
+}
+
+// Min returns the min value, swapping min/max if not ordered
+func (rn *RangedNumber) Min() int {
+	if rn.min > rn.max {
+		rn.Set(rn.max, rn.min)
+	}
+
+	return rn.min
+}
+
+// Max returns the max value, swapping min/max if not ordered
+func (rn *RangedNumber) Max() int {
+	if rn.min > rn.max {
+		rn.Set(rn.max, rn.min)
+	}
+
+	return rn.max
+}
+
+// Set sets the min and max values, ordering the arguments if necessary
+func (rn *RangedNumber) Set(min, max int) *RangedNumber {
+	rn.SetMin(min)
+	rn.SetMax(max)
+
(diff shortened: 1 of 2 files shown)


d2common/d2math/ranged_number.go
```diff
@@ -0,0 +1,101 @@
...
+// Mul multiplies this ranged number by the given ranged number, returning this one
+func (rn *RangedNumber) Mul(other *RangedNumber) *RangedNumber {
+	return rn.Set(rn.min * other.min, rn.max * other.max)
+}
+
+// Div divides this ranged number by the given ranged number, returning this one
+func (rn *RangedNumber) Div(other *RangedNumber) *RangedNumber {
+	return rn.Set(rn.min / other.min, rn.max / other.max)
+}
+
```
malavv, line 93: I'm sure some wouldn't be happy to add something we have no use for yet, but I often see these kind of type with methods : `IsIn(n int)`, `overlaps(rng RangedNumber)`, `encloses(rng RangedNumber)`. Potentially for another PR down the line.

d2common/d2math/ranged_number.go
```diff
@@ -0,0 +1,101 @@
...
+// RangedNumber is a number with a min and max range
+type RangedNumber struct {
+	min int
+	max int
+}
+
+// Min returns the min value, swapping min/max if not ordered
+func (rn *RangedNumber) Min() int {
+	if rn.min > rn.max {
```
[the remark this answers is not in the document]
gravestench: I'm not sure, it was worrying me that they could be out of order so I added this, but you're probably right.

d2common/d2math/ranged_number.go
gravestench: This was originally on my feature branch for stats, but as I finished it I figured I would push now (to keep the upcoming stat PR small)

gravestench:
**Should these be immutable?** 

All of the methods currently alter the struct. I have a hunch that this is probably a bad thing, and that it would be easier to use if the methods returned altered clones.

malavv:
> **Should these be immutable?**
> 
> All of the methods currently alter the struct. I have a hunch that this is probably a bad thing, and that it would be easier to use if the methods returned altered clones.

Good point, I can think of good points for both so I don't have any specific preference. What you've done sounds good, and might be more efficient down the line memory wise.

PR #582, opened 2020-07-13
Add static checks to help ensure struct implementations conform to interfaces

sunhay:
A fairly common pattern that I've seen - this removes a class of issues with respect to runtime interface checks, but will also make it easier to confirm implementations are conforming to interfaces (i.e fail fast).

https://golang.org/doc/effective_go.html#blank_implements


gravestench:
This is really cool, thanks for the PR! I've been trying to think of a better way of making sure things implement the interfaces.

PR #590, opened 2020-07-14
Stat descriptions + tests, Skilldesc.txt loader

gravestench:
- adding call to skilldesc loader in app.go
- adding skilldesc resource path
- adding skilldesc loader in d2datadict
- monstats field name was incorrect, discovered while writing unit tests for stats descriptions
- moved stat description methods out of `ItemStatCostRecord` and into `Stat`
- wrote some simple unit tests for checking stat descriptions (these need work, and so do the stat description functions!)
- npc needed an edit for the corrected fieldname from monstats records


gravestench:
It's worth noting that the stat description tests, for all stats, are currently passing.

Also, the build for this is failing because the runner doesn't have a copy of the MPQ's. Not sure about what to do about that other than make a ton of mock records/string tables (which would be a lot of work).

@essial is there any way we can get the MPQ's on the build platform?

gravestench:
Here's a copy of the output from the tests: https://pastebin.com/raw/VtSsMbgJ

gravestench:
`d2common.TranslateString` returns the table string key if the table lookup fails, and _the tests currently abuse this behavior_. The mock data dictionary records in the tests contain the table string entries (instead of the keys which would normally be used), and this ends up being functionally equivalent as far as stat descriptions are concerned.

PR #592, opened 2020-07-14
rely on App to know how to navigate between screens using a callback …

carrelld:
…interface with explicit methods

remove unused struct vars that were only stored in order to pass to the next screens

consolidate create game code to single method

export ScreenMode consts and SetScreenMode method to allow app to create the correct screens

I decided against a `sceneManager.SetNextScene(d2enum.SceneMainMenu)` method for a couple reasons. With explicit methods, we can keep the method smaller and easier to understand. Secondly, we can define what parameters are needed for the specific action.

### DIFF
--- a/d2app/app.go
+++ b/d2app/app.go
@@ -29,6 +29,8 @@ import (
 	"github.com/OpenDiablo2/OpenDiablo2/d2core/d2screen"
 	"github.com/OpenDiablo2/OpenDiablo2/d2core/d2ui"
 	"github.com/OpenDiablo2/OpenDiablo2/d2game/d2gamescreen"
+	"github.com/OpenDiablo2/OpenDiablo2/d2networking/d2client"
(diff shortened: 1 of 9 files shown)


d2app/app.go
```diff
@@ -601,3 +603,39 @@ func updateInitError(target d2interface.Surface) error {
 
 	return nil
 }
+
+func (a *App) ToMainMenu() {
```
essial, line 607: Can we turn this into a TransitionScene(d2enum.SceneType) or something?

carrelld: >I decided against a sceneManager.SetNextScene(d2enum.SceneMainMenu) method for a couple reasons. With explicit methods, we can keep the method smaller and easier to understand. Secondly, we can define what parameters are needed for the specific action.

carrelld: Some of the transitions may require a `filePath` or `connectionHost` or `connectionType` or a myriad of other arguments. This would be difficult to pass these cleanly in a generic transition interface.

PR #594, opened 2020-07-16
Vector benchmark

danhale-git:
This PR is almost all test and benchmark changes (vector_test.go).

It adds benchmarks for the remaining (now all) functions in vector_test.go. Some of these might not be quite right. My next step is to review each benchmark and optimise its corresponding Vector method.

### DIFF
--- a/d2common/d2math/d2vector/vector_test.go
+++ b/d2common/d2math/d2vector/vector_test.go
@@ -607,36 +607,59 @@ func BenchmarkVector_ReflectSurface(b *testing.B) {
 	}
 }
 
-func TestRotate(t *testing.T) {
-	up := NewVector(0, 1)
+func TestVector_Rotate(t *testing.T) {
 	right := NewVector(1, 0)
-
 	c := right.Clone()
+
+	up := NewVector(0, 1)
 	angle := -up.SignedAngle(right)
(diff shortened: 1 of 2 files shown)


d2common/d2math/d2vector/vector_test.go
```diff
@@ -7,50 +7,17 @@ import (
...
-		t.Errorf("%s: wanted %f: got %f", description, want, got)
-	}
-}
-
-func evaluateChanged(description string, original, clone Vector, t *testing.T) {
-	if !original.Equals(clone) {
-		t.Errorf("%s: changed vector %s to %s unexpectedly", description, clone, original)
-	}
-}
+//nolint:gochecknoglobals // These variables are assigned to in benchmark functions to avoid compiler optimisations
```
danhale-git, line 10: I initially ran into what I assume is the compiler optimisation issue and assigning to a package level variable does resolve it. It doesn't seem to be necessary when operating on the vector itself, only when returning a value which is not used. Assigning to _ does not work.

I don't think it's possible to avoid the unused variable warnings without either adding an unused function which assigns them to something (no) or some sort of over the top hackery. 

Happy to hear about any better alternatives to all this. Since it's in a _test.go file I guess it isn't really the end of the world.

PR #599, opened 2020-07-18
delint_d2networking

gravestench:
edits in d2core and d2game:
- remove a struct copy and instead pass pointers for `HeroStatsState` and `CharacterEquipment` structs
- change `Id` to `ID` in struct fields and function arguments

There are still some lint errors for the switch statements that handle the packet types, as the functions are long. Not sure if I should add a nolint, but let me know if I should, I don't see much way around it.

There is also a lint error for `io.Copy` calls that warn about a gzip compression bomb, but not sure what to do about it.
image

### DIFF
--- a/d2game/d2gamescreen/game.go
+++ b/d2game/d2gamescreen/game.go
@@ -153,7 +153,7 @@ func (v *Game) Advance(tickTime float64) error {
 
 func (v *Game) bindGameControls() {
 	for _, player := range v.gameClient.Players {
-		if player.Id != v.gameClient.PlayerId {
+		if player.Id != v.gameClient.PlayerID {
 			continue
 		}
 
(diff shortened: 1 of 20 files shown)


d2game/d2gamescreen/game.go
```diff
@@ -153,7 +153,7 @@ func (v *Game) Advance(tickTime float64) error {
 
 func (v *Game) bindGameControls() {
 	for _, player := range v.gameClient.Players {
-		if player.Id != v.gameClient.PlayerId {
```
Iniquitatis, line 156: One more `Id` here.

gravestench: thank you!

gravestench: I'll put this in a PR for d2game lint errors

PR #605, opened 2020-07-18
Implemented Position in Path and PathTile

danhale-git:
This PR implements Vector in Path and PathTile. This will eventually enable the removal of magic numbers.

### DIFF
--- a/d2core/d2map/d2mapengine/walk_mesh.go
+++ b/d2core/d2map/d2mapengine/walk_mesh.go
@@ -3,6 +3,8 @@ package d2mapengine
 import (
 	"math"
 
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2math/d2vector"
+
 	"github.com/OpenDiablo2/OpenDiablo2/d2common"
 	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2astar"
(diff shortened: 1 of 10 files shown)


d2core/d2map/d2mapengine/walk_mesh.go
```diff
@@ -51,8 +53,9 @@ func (m *MapEngine) RegenerateWalkPaths() {
 			index := subTileX + (subTileY * m.size.Width * 5)
```
gravestench, line 53: i think we should make `numSubtilesPerTile` a constant in d2common

danhale-git: Yes it is. This PR is the minimum code needed to implement Position. There will be another PR at some point to actually remove all the magic numbers. Possibly some more Position implementation before that.

The purpose of Position is to handle the whole tile/subtile thing. Rather than use a constant here, `subTileX` and `subTileY` will be a `Position`. This way nothing except Position has to know about the tile to subtile ratio. Note, this PR also adds NewPositionTile which creates a position from tile coordinates, again using `Position.subTilesPerTile`.

gravestench: That makes way more sense than using a subtiles constant everywhere, i like it.

PR #608, opened 2020-07-19
Theta* pathfinding

Ziemas:
This implements theta* pathfinding. This is basically a* but when searching for paths we straighten them out by testing if the parent node has vision to the new neighbor node and skipping the current node if they do.

It is currently slightly buggy for reasons unkown to me and will sometimes happily run through unwalkable tiles.

### DIFF
--- a/d2common/path_tile.go
+++ b/d2common/path_tile.go
@@ -13,6 +13,7 @@ type PathTile struct {
 	UpLeft, UpRight     *PathTile
 	DownLeft, DownRight *PathTile
 	Position            d2vector.Position
+	SubX, SubY          int
 }
 
 // PathNeighbors returns the direct neighboring nodes of this node which can be pathed to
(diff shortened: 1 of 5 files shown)


essial:
There's broken tests in this PR.

d2common/path_tile.go
```diff
@@ -13,6 +13,7 @@ type PathTile struct {
 	UpLeft, UpRight     *PathTile
 	DownLeft, DownRight *PathTile
 	Position            d2vector.Position
+	SubX, SubY          int
```
danhale-git, line 16: `SubX` and `SubY` are equal to `Position.Floor().X()` and `Y()` (or maybe that vector+ 1,1 - see `Position.RenderOffset`).

`Position` used to contain methods for these things but I removed them as they were unused. Since the sub tile position is used here, I suppose we should add a `Position.SubTile` method to get that value.

PR #609, opened 2020-07-20
entity debug rendering

gravestench:
needed to add getter method for position/velocity to MapEntity interface, so that's the reasoning for changing objects/npcs/player/etc

this currently brings allocs/s up to 10 (you can check by running `fps` command, se we need to look into d2debugutils and how it handles drawing text. either that, or at line `d2maprenderer/renderer.go:336` we need to return if the entity is not on-screen. I tried to do it, but gave up after about an hour.

### DIFF
--- a/d2common/d2interface/map_entity.go
+++ b/d2common/d2interface/map_entity.go
@@ -1,10 +1,13 @@
 package d2interface
 
+import "github.com/OpenDiablo2/OpenDiablo2/d2common/d2math/d2vector"
+
 // MapEntity is something that can be positioned on and rendered on the game map
 type MapEntity interface {
 	Render(target Surface)
 	Advance(tickTime float64)
-	GetPosition() (float64, float64)
+	GetPosition() d2vector.Position
+	GetVelocity() d2vector.Vector
 	GetLayer() int
 	GetPositionF() (float64, float64)
 	Name() string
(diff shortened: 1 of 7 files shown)


danhale-git:
This fixes the zero vector issue:
```
func (m *mapEntity) Step(tickTime float64) {
	if m.atTarget() && !m.hasPath() {
		if m.done != nil {
			m.done()
			m.done = nil
		}

		m.velocity.SetLength(0)

		return
	}

	m.setVelocity(tickTime * m.Speed)

	v := m.velocity.Clone() // Create a new vector

	for {
		applyVelocity(&m.Position.Vector, &v, &m.Target.Vector) // Pass the new vector to the function which alters it

		if m.atTarget() {
			m.nextPath()
		}

		if v.IsZero() { // Check if the new vector is zero (keeping this as m.vector.IsZero() would break the test (infinite loop)
			break
		}
	}
}
```

danhale-git:
When graphing DrawLine(m.Position, m.Position.Add(m.velocity), dividing m.velocity by ticktime (DivideScalar) would get you the velocity in sub tiles per second. But you might as well just use an arbitrary number to make your line long enough to see, or assume the tick time is 0.05 seconds which it seems to be most of the time, roughly.

Issue #504, opened 2020-06-30
Remove all global variables and singletons


essial:
We need to remove all global references, which include all singletons.

Please make sure you track on this spreadsheet so that we do not ninja each other's changes:
https://docs.google.com/spreadsheets/d/1y6S9-rAB3MMIofSc09RoO9eHGAv76FmmTeHqYjcLzz0/edit#gid=1196479007

gravestench:
Just an update, work towards this issue:
#696 
#735 
#738 
#742

We still need to remove singletons in:
-  `d2networking/d2server/game_server.go`
- `d2common/d2util/debug_print.go`
- `d2common/d2fileformats/d2tbl/text_dictionary.go`
- `/d2core/d2config/d2config.go`

gravestench:
#824 resolves the game server

gravestench:
#900 resolves the text dictionary singleton

gravestench:
#899 resolves d2config singleton

gravestench:
After #899, #900, and #901 get merged we can probably close this issue.

There are still some things like this in the project, but not many.
```golang
var (
	errPlayerAlreadyExists = errors.New("player already exists")
	errServerFull          = errors.New("server full") // Server currently at maximum TCP connections
)
```

PR #611, opened 2020-07-22
Removed some global variables

gurkankaymak:
related to #504


gravestench:
This is breaking the build, try to run `go test ./...`

PR #613, opened 2020-07-22
Position implemented in newMapEntity.

danhale-git:
This PR makes the use of pointers consistent in `Vector` and implements `Position` in newMapEntity.
Odd mix of changes I'm afraid but they are all straightforward.


danhale-git:
I can make NewVector and Clone return pointers. I do still need to learn how to properly profile Vector and Position, now that they are being used.

I'll make the change tomorrow morning in about 10 hours.

danhale-git:
I will do that in a new branch. Feel free to merge this if it's good.

essial:
Merge conflicts. Also I agree we should return pointers so we can chain them.

danhale-git:
This should be two PRs. I'm going to rewrite from a fresh branch and submit the Vector changes separately. I will change all the method return values to pointers this time.

PR #620, opened 2020-07-25
fix(camera): focus the local player on viewport init fix #612

### LINKED ISSUE #612: Camera should start at player position
When entering a game the viewport camera starts at world position (0, 0) and moves to the player.

Instead, it should be immediately moved to the player position.

Benzhaomin:
not the cleanest but keeps it simple and without the renderer having to know about Player stuff


gravestench:
good work!

Issue #634, opened 2020-07-28
Write remaining d2datadict loaders


andrewDoing:
**Problem**
Not all the *.txt files are loaded into the game from /data/global/excel

**Solution**
Write data loaders for the remaining .txt files

**Files remaining:**
patch_d2.mpq
- [x] books.txt
- [ ] compcode.txt (Hardcoded Reference File)
- [x] ElemTypes.txt
- [ ] events.txt (Hardcoded Reference File)
- [x] itemratio.txt
- [x] misscalc.txt
- [ ] monai.txt (Hardcoded Reference File)
- [x] monequip.txt
- [x] MonLvl.txt
- [x] MonMode.txt
- [ ] MonPlace.txt (Hardcoded Reference File)
- [x] MonPreset.txt (not sure, monpreset.txt exists somewhere)
- [x] MonProp.txt
- [x] monseq.txt
- [x] monsounds.txt
- [x] MonType.txt
- [x] monumod.txt
- [x] npc.txt
- [x] Overlay.txt
- [x] pettype.txt
- [x] PlrMode.txt
- [x] Runes.txt
- [x] shrines.txt
- [x] skillcalc.txt
- [x] soundenviron.txt
- [x] UniqueAppellation.txt
- ~~UniqueTitle.txt~~

d2exp.mpq
- ~~Aiparms.txt~~
- ~~Arena.txt~~
- [ ] ArmType.txt
- [x] AutoMap.txt
- ~~CharTemplate.txt~~
- [ ] Composit.txt
- [ ] HitClass.txt (Hardcoded Reference File)
- [x] LvlWarp.txt
- ~~MonItemPercent.txt~~
- ~~MonName.txt~~
- [x] ObjMode.txt (Hardcoded Reference File)
- [x] ObjType.txt
- [x] PlayerClass.txt
- [ ] PlrType.txt (Controls the animation used by characters
- [ ] RarePrefix.txt (in resource_paths.go, but doesn't have its own loader)
- [ ] RareSuffix.txt (in resource_paths.go, but doesn't have its own loader)
- [x] SoundEnviron.txt (lowercase named file in patch_d2.mpq, probably the same?)
- [x] Sounds.txt
- [ ] StorePage.txt (Hardcoded Reference File)
- [x] TreasureClass.txt
- [x] WeaponClass.txt
- [x] automagic.txt
- [ ] belts.txt
- [x] bodylocs.txt
[...]

PR #641, opened 2020-07-29
d2datadict: Add UniqueTitle loader

andrewDoing:
Issue #634


andrewDoing:
Looking at this again, it seems like this is overkill considering the file is just one column of UniqueTitles. Should I still use a dict or is there a simpler solution that will work?

Ziemas:
I think this txt file is actually unused (as per https://d2mods.info/forum/viewtopic.php?t=34455)

andrewDoing:
Closed because the file is unused

PR #643, opened 2020-07-29
Replace globals with initializers (d2gui d2input)

AndrejMijic:
Issue #504


essial:
Have you profiled memory usage on this to make sure we aren't burning allocs with returning of these maps?

AndrejMijic:
The ebiten_input functions seem to increase malloc time, so I've reverted the change.

Global map:
[ebiten_input_global.pdf](https://github.com/OpenDiablo2/OpenDiablo2/files/4997803/ebiten_input_global.pdf)

Function:
[ebiten_input_func.pdf](https://github.com/OpenDiablo2/OpenDiablo2/files/4997804/ebiten_input_func.pdf)

PR #648, opened 2020-07-30
Add pettype.txt loader

AndrejMijic:
Issue #634


essial:
Make sure you put your name in the credits, you're knocking these out like crazy!

PR #666, opened 2020-08-01
Minor edits: debug entity frame bounds, debug `spawnmon` command

gravestench:
Adding debug printing of entity frame bounds. This is currently buggy as render offset for map entities is not center of the sprite.

added `spawnmon` command for spawning monsters.


essial:
This is the best PR.

PR #675, opened 2020-08-04
Add monpreset.txt loader

AndrejMijic:
Issue #634 

This should conclude all patch_d2.mpq .txt files.


AndrejMijic:
I managed to run the correct linter, it did not complain. Perhaps the map should be to pointers? I just noticed it.

PR #694, opened 2020-08-05
D2mapengine remove entity, minor edits

gravestench:
- imlemented removeEntity (missiles now dissappear)
- added rgba color provider function to d2common
- fixed some of the lint errors in d2map


essial:
Test failed fyi

gravestench:
already looking at it, i always forget to check the tests

gravestench:
It becomes apparent that we need to handle depth sorting after storing map entities in a map (instead of a slice). We would still have to look into this even if they weren't stored in a map, but the issue is obvious now.
ezgif-4-8c4a3e0ef782

PR #698, opened 2020-08-06
Networking Refactor

stephenhoran:
Couple of big items here...

- Game Server is **almost** no longer a singleton... need to review local connections. A side effect of this is running the networking stack separately is now very easy.
- Refactor to TCP Connections.
- Game Server now concurrently handles each connection.
- Connections are now handled using the net.Conn interface, in the event we need to change between TCP/UDP/Something else it should be plug and play.
- Context implemented into the Game Server, however needs to be better used.
- A Packet manager manages routing of packets to each connection. Each packet received is passing on a channel to the packet manager, so each packet is handled FIFO.
- The top level packet has been modified slightly, the data component is no longer an interface{}, but a json.RawMessage ([]byte). This allows for unmarshaling of sub message types to be much cleaner.
- Compression has been removed, for now. So far all of our packets are much smaller then the MTU size, so compression is not saving us any packets. Unless packets which are sent very frequently exceed the MTU size compression might actually be costing us.
- Switched over to using the JSON Decoder which will stream bytes from each connection much cleaner then managing our own bytes and seeking to new lines.
- Some helpers have been added to each packet.

Still to do...

- Documentation and better code comments.
- Handle ALL  of the current packet types. Only a handful implemented.
- Connection cleanup, currently the server does not properly remove dead connections.
- Create a CLI tool for running the networking stack in a dedicated way.
- TESTING. Need to really focus on developing a way to test before moving forward.
- Address all the TODO's.
- Evaluate how we do local connections, after this, DRY it up.


d2networking/d2client/d2remoteclient/remote_client_connection.go
```diff
@@ -24,7 +21,7 @@ import (
 type RemoteClientConnection struct {
 	clientListener d2networking.ClientListener // The GameClient
 	uniqueID       string                      // Unique ID generated on construction
-	udpConnection  *net.UDPConn                // UDP connection to the server
+	tcpConnection  *net.TCPConn                // UDP connection to the server
```
gravestench, line 24: this is minor, the comment needs to be updated

d2networking/d2client/game_client.go
```diff
@@ -189,7 +204,11 @@ func (g *GameClient) handleSpawnItemPacket(packet d2netpacket.NetPacket) error {
 }
 
 func (g *GameClient) handleMovePlayerPacket(packet d2netpacket.NetPacket) error {
-	movePlayer := packet.PacketData.(d2netpacket.MovePlayerPacket)
+	movePlayer, err := d2netpacket.UnmarshalMovePlayer(packet.PacketData)
+	if err != nil {
+		return err
+	}
+
 	player := g.Players[movePlayer.PlayerID]
```
gravestench, line 212: also minor, but we may want to check that `player` exists in these kinds of cases

d2networking/d2netpacket/d2netpackettype/message_type.go
```diff
@@ -25,6 +27,8 @@ const (
 	ServerClosed                                         // Sent by the local host when it has closed the server
 	CastSkill                                            // Sent by client or server, indicates entity casting skill
 	SpawnItem                                            // Sent by server
+
+	UnknownPacketType = 666
```
gravestench, line 31: a20ddd95ed33c6937246c1f37232e519

d2networking/d2netpacket/net_packet.go
```diff
@@ -1,12 +1,42 @@
 package d2netpacket
 
-import "github.com/OpenDiablo2/OpenDiablo2/d2networking/d2netpacket/d2netpackettype"
+import (
+	"encoding/json"
+	"github.com/OpenDiablo2/OpenDiablo2/d2networking/d2netpacket/d2netpackettype"
```
gravestench, line 5: lint error

d2networking/d2netpacket/packet_add_player.go
```diff
@@ -24,16 +25,28 @@ type AddPlayerPacket struct {
 // AddPlayerPacket with the data in given parameters.
 func CreateAddPlayerPacket(id, name string, x, y int, heroType d2enum.Hero,
```
gravestench, line 26: also minor, I think we are trying to standardize our factory function names as `NewXYZ` instead of `CreateXYZ`

d2networking/d2server/d2tcpclientconnection/tcp_client_connection.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"encoding/json"
+	"github.com/OpenDiablo2/OpenDiablo2/d2networking/d2client/d2clientconnectiontype"
+	"github.com/OpenDiablo2/OpenDiablo2/d2networking/d2netpacket"
+	"net"
+
+	"github.com/OpenDiablo2/OpenDiablo2/d2game/d2player"
+)
+
+type TCPClientConnection struct {
```
gravestench, line 12: `TCPClientConnection` and it's methods need comments

PR #699, opened 2020-08-06
d2ui refactor

gravestench:
- changed `UI` struct to `UIManager`, added provider to create an instance, removed the singleton
- UI element provider functions have been converted to methods of the UI Manager
- Screens now use the UI manager to create UI elements
- game panels in d2player now use the UI Manager to create UI elements
- only the UI manager knows about "widgets"; calls to `d2ui.AddWidget` in Screen instances have been removed
- changed ui element provider methods from `Create` to `New`


gravestench:
This PR brings the lint error count for d2ui from 93 down to 4

Issue #689, opened 2020-08-05
Implement CalcStrings


gravestench:
txt files can have static or calculated values.

here's an example of a calcstring (from skills.txt, column `EDmgSymPerCalc`):
```
(skill('Lightning Strike'.blvl)+skill('Lightning Bolt'.blvl)+skill('Charged Strike'.blvl)+skill('Lightning Fury'.blvl)) * par8
```

here is a fileguide from phrozen keep that describes formulae (calcstrings):
https://d2mods.info/forum/kb/viewarticle?a=371

here is the skills.txt file guide:
https://d2mods.info/forum/kb/viewarticle?a=440

AndrejMijic:
Do we want this parsed into an expression tree which can be evaluated when necessary?

PR #706, opened 2020-08-16
Add initial calculation string parser

AndrejMijic:
Issue #689.

The parser parses the calculation strings into a Calculation type which can be evaluated by calling the Eval function.
The parsing is done immediately when loading.

Currently, property references such as skill("Sacrifice".lvl) etc. are not implemented and always return 1.

References to the current skill (par1, lvl etc.) are automatically expanded with the current skill name.
Calculations for skills are appear in skills and skillsdesc. In skillsdesc, references to the current skill are not fully expanded, as I'm not sure how to elegantly get the skill name.


essial:
Is it possible to add some tests for these so we can check for memory leaks and stuff?

AndrejMijic:
I can write some tests for expressions, but I'm not sure where memory leaks could arise.

essial:
They show up in the strangest of places 😆

PR #714, opened 2020-09-06
D2loader

gravestench:
This is the beginning of an attempt to decouple how we load assets. The big picture here is that we have a `Loader` which works with `Source`s and `Asset`s,

A `Source` can be a filesystem, or an MPQ archive, and we can extend it to include other formats. It's really just a wrapper for the file retrieval logic.
 
An `Asset` is something that implements `io.ReadSeeker`, and has a few other methods of its own. This is really just a wrapper for the file data.

Currently, the `Loader` can have filesystem and MPQ sources, and i've added tests for each.

here's an example of how the loader can be used:
```golang
loader := NewLoader()
loader.SetLogger(os.Stdout, LogLevelDebug)

loader.AddSource("/users/me/.config/OpenDiablo2") // allows us to load our config.json
loader.AddSource(".") // allows us to mimic the `-direct` command line arg of original diablo2

// the load order we are familiar with
loader.AddSource("/some/path/to/patch_d2.mpq")
loader.AddSource("/some/path/to/d2exp.mpq")
loader.AddSource("/some/path/to/d2xmusic.mpq")
loader.AddSource("/some/path/to/d2xtalk.mpq")
loader.AddSource("/some/path/to/d2xvideo.mpq")
loader.AddSource("/some/path/to/d2data.mpq")
loader.AddSource("/some/path/to/d2char.mpq")
loader.AddSource("/some/path/to/d2music.mpq")
loader.AddSource("/some/path/to/d2sfx.mpq")
loader.AddSource("/some/path/to/d2video.mpq")
loader.AddSource("/some/path/to/d2speech.mpq")


config, cfgErr := loader.Load("config.json")
if cfgErr != nil {
    config, _ = loader.Load("config.default.json")
}

// if there is a `data/global/excel/monstats.txt` in the user config dir or the 
// OpenDiablo2 dir this file will be pulled from there
monstats, err := loader.Load("data/global/excel/monstats.txt")
```

None of this is currently being used anywhere in the project, this is just some preliminary work i've done and I want some feedback. I think it's worth noting that although we *can* have a single loader for everything, we don't really need to do that. There can be separate loaders for loading config files, or a loader that is only used for audio/video stuff, etc. I think this leaves the engine flexible enough that you could even have a game server that provides a network source for custom game files.

I'm still trying to figure out how to fit this into the engine in the right way, so thoughts and feedback appreciated.


gravestench:
the failing tests is unrelated and has been addressed in #715

PR #723, opened 2020-09-09
fix glfw error in ci workflow

gravestench:
https://github.com/hajimehoshi/ebiten/issues/939#issuecomment-624277149

thank you @tommyblue


tommyblue:
thanks 😃

PR #726, opened 2020-09-12
remove d2asset singleton

gravestench:


gravestench:
This is part of the fix for issue #504

PR #728, opened 2020-09-13
improve AssetManager implementation

gravestench:
* d2loader.Loader is now used for loading all file assets, asset manager caches higher-level assets like animations
 * removed the individual managers inside of d2asset, only one asset manager
 * AssetManager now has a cache for each category of file (fonts, palettes, animations, string tables)
 * created a type for TextDictionary (the txt file structs) in d2datadict
 * fixed a file path bug in d2loader Source
 * fixed a asset stream bug in d2loader Asset
 * d2loader.Loader now needs a d2config.Config on creation (for resolving locale files)
 * updated the mpq file in d2loader/testdata, added test case for "sub-directory"
 * added a Data method to d2asset.Asset. The data is cached on first full read.
 * renamed ArchiveDataStream to DataStream in d2interface
 * moved palette utility func out of d2asset and into d2util
 * d2loader bugfix for MacOS mpq load issue (mpq's dont have file extensions on macOS!)
 * removed the palette, font, animation, file, and archive manager interfaces from d2interface


gravestench:
Circle-CI build seems to be broken
image

d2common/d2loader/filesystem/source.go
```diff
@@ -43,7 +43,7 @@ func (s *Source) fullPath(subPath string) string {
 	return filepath.Clean(filepath.Join(s.Root, subPath))
 }
 
-// Path returns the Root dir of this file system source
+// String returns the Root dir of this file system source
```
presiyan-ivanov, line 46: Can you comment why this was changed? I suppose you replaced this by mistake?

d2common/d2loader/mpq/asset.go
```diff
@@ -35,14 +42,58 @@ func (a *Asset) Path() string {
...
+
+	return totalRead, nil
 }
 
 // Seek will seek the read position for the next read operation
 func (a *Asset) Seek(offset int64, whence int) (n int64, err error) {
 	return a.stream.Seek(offset, whence)
 }
 
+// Seek will seek the read position for the next read operation
```
presiyan-ivanov, line 58: You forgot to change the doc

d2common/d2loader/mpq/asset.go
```diff
@@ -35,14 +42,58 @@ func (a *Asset) Path() string {
...
+func (a *Asset) Close() (err error) {
+	_, err = a.Seek(0, 0)
```
presiyan-ivanov, line 60: I think this should be `a.stream.Close()`

d2common/d2loader/filesystem/source.go
gravestench: the change to the comment is a mistake

d2common/d2loader/mpq/asset.go
gravestench: i will update in a moment

d2common/d2loader/mpq/asset.go
gravestench: So this is a strange issue i was having. Usually, calling close will set a stream to nil, and if i call the mpq stream `Close` method it will do the same. But, because i am caching these mpq file entries for retrieval later, i don't close the stream, i just seek to 0 and keep it open. The next time this entry is retrieved from the cache, it will be ready to read from. I think the only time it will be closed is after it is ejected from the cache and gc gets to it.

d2common/d2loader/filesystem/source.go
gravestench: I should note that I chose to have separate `String` and `Path` methods because I wanted it to implement `fmt.Stringer`, so if we use this object in something like `fmt.Sprintf` it will print the path of the source file.

presiyan-ivanov (approved):

d2common/d2loader/mpq/asset.go
gravestench: The asset needs to implement `io.Reader`, `io.Writer`, and `io.Seeker`, an unexported `reset` method would just be a wrapper for `Seek(0, 0)`. I've opted to just add a comment here that explains what the magic is.

PR #729, opened 2020-09-13
Prevent player movement and casting if current spell cast is not fini…

presiyan-ivanov:
…shed yet


gravestench (approved):
Looks good to me

PR #733, opened 2020-09-17
Example of a data loader using reflection.

danhale-git:
This PR is intended as an example for something I discussed with @gravestench. Not sure why it can't be automatically merged but it probably shouldn't be merged at all so no biggie.

The PopulateStruct function works as a 'generic' way to fill a struct with some data in a map. The struct must be defined and passed to reflect.ValueOf(), but no specific code is required to fill that struct. The return value of `reflect.ValueOf()` gives us the parameter `values reflect.Value` which represents any struct we want to populate, as something resembling a map of field names and values. The only code that must run for the specific type is:

`record := BooksRecord{}`
`err := d.PopulateStruct(reflect.ValueOf(&record).Elem())`
`Books[record.Namco] = &record`

There is another implementation [here](https://github.com/danhale-git/runrdp/blob/2293cdccf7d7eff021dbb34ac27e898848070a2e/internal/config/config.go#L184) which covers more types.

That implementation was my first attempt at using reflection in Go. I don't know about the performance implications of using it for all data loading, there probably are some.

I've just quickly shoe-horned this into the existing books loader. No doubt it could be used to clean up the code much more than I have, if it is viable.

### DIFF
--- a/d2common/d2data/d2datadict/books.go
+++ b/d2common/d2data/d2datadict/books.go
@@ -1,7 +1,9 @@
 package d2datadict
 
 import (
+	"fmt"
 	"log"
+	"reflect"
 
 	"github.com/OpenDiablo2/OpenDiablo2/d2common"
 )
@@ -13,7 +15,7 @@ type BooksRecord struct {
 	Completed       string
 	ScrollSpellCode string
 	BookSpellCode   string
-	pSpell          int
+	PSpell          int
 	SpellIcon       int
 	ScrollSkill     string
 	BookSkill       string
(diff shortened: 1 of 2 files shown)


danhale-git:
I should also note that books.go is a nice simple case where all the struct field names pretty much match the data field names and everything is a string or an int. Things could potentially get much more tricky with other data loaders.

gravestench:
@danhale-git is this safe to close for now?

danhale-git:
Yes absolutely.

gravestench:
I think that this is the right approach, but I don't think i'm the right person to implement it. 

We just did a lot of work removing `d2datadict` and all its singletons, it may be the right time to try and implement this.

PR #736, opened 2020-09-19
Added RecordManager implementation to remove d2datadict singletons

gravestench:
The PR offers a solution to the 68 data dictionary singletons in `d2common/d2data/d2datadict`. It provides a `RecordManager` implementation, which handles loading all data dictionary files and exporting them. `d2core/d2records` aims to fully replace `d2common/d2data/d2datadict`.

### Why is this PR so huge?
90% of what is in this PR is actually just what we currently have for all 68 of the data dictionary loaders, but they have been reformatted to adhere to this function signature:
```golang
type dataLoader func(r *RecordManager, d *d2txt.DataDictionary) error
```

Another reason why this commit is so large is because all of the loader and struct declarations have been separated into their own files.

### What is the record manager?
The `RecordManager` is an exported member of `d2asset.AssetManager`. The `RecordManager` is a struct which exports all of the data dictionary records (what are currently the singletons).

Here is some of what the `RecordManager` exports:
```golang
// RecordManager stores all of the records loaded from txt files
type RecordManager struct {
	BodyLocations
	Calculation struct {
		Skills   Calculations
		Missiles Calculations
	}
	Character struct {
		Classes PlayerClasses
		Events
		Experience ExperienceBreakpoints
		MaxLevel   ExperienceMaxLevels
		Modes      PlayerModes
		Stats      CharStats
	}
	ComponentCodes
	DifficultyLevels
	ElemTypes
	Hirelings
	Item struct {
		All CommonItems

		Armors  CommonItems
		Misc    CommonItems
		Weapons CommonItems

		Equivalency         ItemEquivalenceMap     
		EquivalenceByRecord ItemEquivalenceByRecord
        /* ... */
        }
}
```

### How does the RecordManager work?
When the provider function `d2asset.NewAssetManager` is called, part of the initialization for the asset manager is to create a `d2loader.Loader` instance, as well as a `d2records.RecordManager` instance. The record manager is created by calling `d2record.NewRecordManager` which calls an un-exported `init` method to establish a mapping of d2resource paths to their appropriate loader functions. However, this only establishes a mapping and does not immediately load anything. The only export that is populated during the `RecordManager` init is the ObjectLookup data which is hard-coded.

After the `RecordManager` init method returns, the `AssetManager` continues initialization. At this point, the asset manager iterates over all data dict paths, passing the path to its own `LoadRecords` method. This is what ends up loading the data dictionaries and makes them available through the `RecordManager` exports.

### How do we remove all of the singletons with this?
any reference to a singleton variable in `d2datadict` needs to be replaced which a reference to an `AssetManager` instance.

This ...
```golang
foo := d2datadict.LevelDetails[0]
```
... should become something like this:
``` golang
foo := assetManager.Records.Level.Details[0]
```


essial:
Good job man. I didn't look through each and every line of code, but the file paths make sense for the changes.

PR #738, opened 2020-09-19
Removing d2datadict singletons

gravestench:
- removed loader calls from d2app
- removed the HeroObjects singleton from `d2core/d2inventory`
- added an InventoryItemFactory in d2inventory
- package-level functions that use data records are now methods of the InventoryItemFactory
- renamed ItemGenerator in d2item to ItemFactory
- package-level functions that use records are now methods of ItemFactory
- d2map.MapEntityFactory now has an item factory instance for creating items
- fixed a bug in unique item record loader where it loaded an empty record
- added a PlayerStateFactory for creating a player state (uses the asset manager)
- updated the test inventory/equipment code in d2player to handle errors from the ItemFactory
- character select and character creation screens have a player state and inventory item factory
- updated item tests to use the item factory


gravestench:
After doing a rebase with the skills stuff that @gurkankaymak , I ended up having to move merge `d2game/d2player/d2player_state_factory` into `d2core/d2hero`, and then make a bunch of edits so that when skills are created they have an asset manager to pull records from.

PR #751, opened 2020-09-24
Updated panel strings to pull from string table

leahiel:
Fixes #749 by pulling from string table instead of hardcoding values in.  Hotkey notifications are still missing.
Thanks for @gravestench for the string table help yesterday.

Also snuck in the Run/Walk tooltip which probably should have been it's own commit.


leahiel:
I need to fix a minor bug that this implements, and add more nuance to the run/walk tooltip.

PR #752, opened 2020-09-24
Updated panel strings to pull from string table v2

leahiel:
Fixes #749 by pulling from string table instead of hardcoding values in. Hotkey notifications are still missing.
Thanks for @gravestench for the string table help yesterday.

Also snuck in the Run/Walk tooltip which probably should have been it's own commit.

Also doesn't have the minor issues that #751 had.

### DIFF
--- a/d2game/d2player/game_controls.go
+++ b/d2game/d2player/game_controls.go
@@ -6,8 +6,10 @@ import (
 	"image/color"
 	"log"
 	"math"
+	"strings"
 	"time"
 
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2fileformats/d2tbl"
 	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2geom"
 	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2util"
 	"github.com/OpenDiablo2/OpenDiablo2/d2game/d2player/help"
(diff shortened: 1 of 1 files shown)


d2game/d2player/game_controls.go
```diff
@@ -848,77 +860,104 @@ func (g *GameControls) Render(target d2interface.Surface) error {
...
+		g.nameLabel.SetPosition(263, 563)
+		g.nameLabel.Render(target)
+	}
+
+	// Create and format Experience string from string lookup table.
+	fmtExp := d2tbl.TranslateString("panelexp")
+	// The English string for "panelexp" is "Experience: %u / %u", however %u doesn't translate well. So
+	// we need to rewrite %u into a formatable Go verb. %d is used in other strings, so we go with that,
+	// keeping in mind that %u likely referred to an unsigned integer.
+	fmtExp = strings.Replace(fmtExp, "%u", "%d", -1)
```
leahiel, line 954: I'm not convinced this was the best way to do this.

gravestench: I think this creates a lint error. Use `strings.ReplaceAll` instead.

leahiel: Fixed in a commit

PR #760, opened 2020-10-06
Introductory notes for OS X

joshjordan:
The process for getting started as a contributor on OS X is a bit confusing. This PR takes a tiny stab at that problem.

- At first, I saw that `build.sh` had no section for darwin, which originally led me to believe it would not run on OS X at all and I'd have a big slog ahead of me to even get started. This turned out not to be the case, but I almost didn't even try. I'm not sure if I already have the dependencies or if they come on default on OS X. I certainly already had `go` setup so I did not bother with the `go_install` section yet.
- There are references to `config.json`, but this file doesn't actually exist until you try to run OD2. And even then, it isn't in the root directory like the docs claim (at least on OS X). Because I'm using Bootcamp to load MPQ files from my Windows partition, I had to edit this file. Updated the docs to better reference the current state of config.json

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -33,7 +33,7 @@ go_install(){
 	  	sudo tar -C /usr/local -xzf go*.linux-amd64.tar.gz > /dev/null 2>&1
       echo "Clean unneeded files"
 	  	rm go*.linux-amd64.tar.gz
-      
+
     elif [ "$distribution" = "Fedora" ]; then
       echo "Downloading Go"
       wget https://dl.google.com/go/go"$go_version".linux-amd64.tar.gz > /dev/null 2>&1
@@ -41,7 +41,7 @@ go_install(){
 	    sudo tar -C /usr/local -xzf go*.linux-amd64.tar.gz > /dev/null 2>&1
       echo "Clean unneeded files"
 	    rm go*.linux-amd64.tar.gz
-    
+
     elif [ "$distribution" = "Debian" ] || [ "$distribution" = "Ubuntu" ] || [ "$distribution" = "Deepin" ]; then
       echo "Downloading Go"
       wget https://dl.google.com/go/go"$go_version".linux-amd64.tar.gz > /dev/null 2>&1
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -71,27 +71,31 @@ fi
...
-    if [ -z "$mesa_detect_arch" ]; then
-		sudo pacman -S libxcursor libxrandr libxinerama libxi mesa libglvnd sdl2 sdl2_mixer sdl2_net alsa-lib --noconfirm > /dev/null 2>&1
-    else
-    sudo pacman -S libxcursor libxrandr libxinerama libxi libglvnd sdl2 sdl2_mixer sdl2_net alsa-lib --noconfirm > /dev/null 2>&1
-    fi
+      if [ -z "$mesa_detect_arch" ]; then
+  		sudo pacman -S libxcursor libxrandr libxinerama libxi mesa libglvnd sdl2 sdl2_mixer sdl2_net alsa-lib --noconfirm > /dev/null 2>&1
+      else
+      sudo pacman -S libxcursor libxrandr libxinerama libxi libglvnd sdl2 sdl2_mixer sdl2_net alsa-lib --noconfirm > /dev/null 2>&1
+      fi
```
joshjordan, line 90: This file mostly contains fixes to inconsistent line endings, but this block should be indented since it is nested within the above `elif`

build.sh
```diff
@@ -71,27 +71,31 @@ fi
...
 
 	  elif [ "$distribution" = "OpenSUSE" ] || [ "$distribution" = "SUSE" ]; then
 		sudo zypper install -y libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel Mesa-libGL-devel alsa-lib-devel libXi-devel > /dev/null 2>&1
 
+    elif [[ "$OSTYPE" == "darwin"* ]]; then
+    # are there dependencies required? did I just have all of them already?
+    echo "Mac OS detected, no dependency installation necessary..."
```
joshjordan, line 97: Should this even be here? I don't particularly like it, but I do want to give a future contributor a clear place to add OS X dependencies without having to figure out *how* to even detect OS X.

essial: For MacOS if Golang is installed, it will simply work. I haven't had to install third party stuff that I know of.

Issue #460, opened 2020-06-25
Pentagrams position in the ingame escape menu


aziule:
**Describe the bug**
When opening the escape menu ingame, the pentagrams are not correctly positioned.

**To Reproduce**
Start a new single player game. Press Escape.
=> The pentagrams appear above the menu.

Click on "OPTIONS"
=> The pentagrams appear above the menu.

=> It basically happens every time a new escape menu layout is loaded.

**Expected behavior**
The pentagrams should:
* Either surround the label hovered by the mouse cursor, if the mouse is hovering an element
* Either surround the topmost label

**Screenshots**
Normal menu:
![image](https://user-images.githubusercontent.com/1004323/85791930-f2531380-b732-11ea-94dc-c348c889b287.png)

With debug mode active (description below about how to activate it):
![image](https://user-images.githubusercontent.com/1004323/85792085-31816480-b733-11ea-867e-291946bfff83.png)

When opening a different layout:
![image](https://user-images.githubusercontent.com/1004323/85792408-bbc9c880-b733-11ea-852a-089a6e10761a.png)

**Additional context**
* It seems to be due to how the `d2gui` pkg works internally (elements position calculation VS rendering). We should someone find a way to calculate the elements positions BEFORE rendering them.
* The debug mode can be activated by going to `d2core/d2gui/layout.go` and uncommenting `if err := l.renderEntryDebug(entry, target); err != nil {`
* Code can be found in `d2game/d2gamescreen/escape_menu.go`
* Pentagrams should only surround `actionableElement` elements and not other regular labels like a title.

essial:
I've dug into this some and made a few adjustments but haven't fixed this yet. What's happening is the layout doesn't seem to be correct at the time we set the option. I've added coded in a merged PR that calls the layout adjust method but that doesn't seem to have fixed the issue.

carrelld:
The `OPTIONS` actionableElement must have been rendered in order for the y offset to be set on the options label itself since the code to calculate its position in first found in render.
```go
func (l *Layout) render(target d2render.Surface) error {
	l.AdjustEntryPlacement()
```

essial:
So it sounds like the bug is that the UI isn't doing a proper layout due to size logic being in the renderer.

PR #763, opened 2020-10-08
Debugging section in readme

joshjordan:
I happened to stumble across #460 which had extremely useful information about debugging. So, I wanted to give us a place to catalog these. I am happy to put these in the wiki instead, but that appears to be untouched for awhile. My opinion is that we should surface them prominently to new developers.

This PR adds this to readme:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ ALL OTHER TRADEMARKS ARE THE PROPERTY OF THEIR RESPECTIVE OWNERS.
 
 ## Status
 
-At the moment (october 2020) the game starts, you can select any character and run around Act1 town. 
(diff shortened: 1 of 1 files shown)


joshjordan:
@essial @gravestench what other debugging tools do we have available? If you can point me at them, I'll describe them. Is there any kind of debug console?

essial:
> @essial @gravestench what other debugging tools do we have available? If you can point me at them, I'll describe them. Is there any kind of debug console?

Press the ``` key and that pops open the terminal. Typing `ls` shows the available commands!

ThomasChr:
Also adding this to the Readme seems to be a good idea to me...

PR #765, opened 2020-10-10
fix(escape_menu): fix pentagrams starting position, fix #460

### LINKED ISSUE #460: Pentagrams position in the ingame escape menu

Benzhaomin:
default menu entry, and rotation direction

feels a bit hackish but it solves the issue right where it is, with no side effect

### DIFF
--- a/d2game/d2player/escape_menu.go
+++ b/d2game/d2player/escape_menu.go
@@ -3,6 +3,7 @@ package d2player
 import (
 	"fmt"
 	"log"
+	"time"
 
 	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2enum"
 	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2interface"
@@ -22,7 +23,7 @@ const (
 	// UI
 	labelGutter    = 10
 	sidePanelsSize = 80
-	pentSize       = 52
+	pentSize       = 54
 	menuSize       = 500
 	spacerWidth    = 10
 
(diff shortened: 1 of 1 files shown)


d2game/d2player/escape_menu.go
```diff
@@ -360,6 +362,7 @@ func (m *EscapeMenu) OnLoad() {
 }
 
 func (m *EscapeMenu) OnEscKey() {
+	// note: original D2 returns straight to the game from however deep in the menu we are
```
joshjordan, line 365: Good catch. I ported this behavior over to `OnEscKey` without noticing this. Want to assign me an issue to fix?

PR #766, opened 2020-10-10
Initial cast overlay implementation. Fix HeroSkill deserialization & …

presiyan-ivanov:
…map entities processing crashing for remote client.

* Casting a skill now plays the corresponding overlay(if any).

* Prevent a crash caused by nil pointer in HeroSkill deserialization, happening when
unmarshalling HeroSkill from packets as a remote client.

* Add PlayerAnimationModeNone to handle some of the Skills(e.g.
Paladin auras) having "" as animation mode.

* Joining a game as remote client now waits for map generation to finish
before rendering map or processing map entities. This is temporary hack to prevent the game from
crashing due to concurrent map read & write exception.

* Send CastSkill packet to other clients.


gravestench (approved):
Looks good to me.

PR #781, opened 2020-10-22
removed a lot of magic number lint errors

gravestench:
all edits in this PR are specifically for removing magic numbers

### DIFF
--- a/d2core/d2asset/animation.go
+++ b/d2core/d2asset/animation.go
@@ -117,20 +117,33 @@ func (a *Animation) Advance(elapsed float64) error {
 	return nil
 }
 
+const (
+	full = 1.0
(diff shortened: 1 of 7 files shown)


ThomasChr:
No magic for you :-(

If you want to correct it, in 'd2map/d2mapentitiy/player.go' on Line 112 and 118 we're using 5 as a magic number:
grafik

This Number has been determined by trying it out and checking if the stamina drain is the same as in d2 with the drain value from the assets. (We stopped the time for a lvl 1 babarian to loose all stamina which is around 25 seconds if i Rememver correctly)

gravestench:
@ThomasChr I didn't edit that file in this PR, it's mostly stuff from the help overlay.

PR #782, opened 2020-10-22
Inital skilltree panel implementation

juander-ux:
Hi,

this PR handles:

* Refactoring of the panel frame into its own class
* Fixing a rendering bug of the inventory for druid/assassin
* Assigning class skills to all characters
* Add an initial skilltree panel where one can click through the tabs and see all the skillicons

foo

Cheers

### DIFF
--- a/d2common/d2enum/hero.go
+++ b/d2common/d2enum/hero.go
@@ -43,3 +43,26 @@ func (h Hero) GetToken() string {
 
 	return ""
 }
+
+// GetToken3 returns a 3 letter token
+func (h Hero) GetToken3() string {
+	switch h {
+	case HeroBarbarian:
+		return "BAR"
+	case HeroNecromancer:
+		return "NEC"
+	case HeroPaladin:
+		return "PAL"
+	case HeroAssassin:
+		return "ASS"
+	case HeroSorceress:
+		return "SOR"
+	case HeroAmazon:
+		return "AMA"
+	case HeroDruid:
+		return "DRU"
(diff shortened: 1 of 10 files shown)


essial:
Great work! If you haven't already, make a PR to add your name to the contributors so it shows up in the credits as well!

gravestench:
Good job, loving how tidy this PR is. 

Consider joining the discord. :)

juander-ux:
Thanks. I just saw that you squashed all my commits into one. Do you do that to all PRs? It takes a considerable amount of time for me to separate my changes into logical steps that contain the why. Squashing them defeats the purpose of me doing this. If you always squash them, then it's no a big deal, I would just like to avoid unnecessary work.

Cheers

gravestench:
@juander-ux I think it's out of habit that the PR's get squashed. I'm sorry about this.

@essial I think we should look out for good PR's like this. @juander-ux clearly invested a lot of time and effort to make his commits coherent, and we've just shit on his hard work. We should not have squashed these.

essial:
Technically speaking, each commit should have been its own PR if they were supposed to be tracked as separate commits on this repo. If we're wanting to do a multi un-squashed PR commit, please let me know in the PR comments as this is not the default mode of operation.

juander-ux:
> If we're wanting to do a multi un-squashed PR commit, please let me know in the PR comments as this is not the default mode of operation.

Sounds good to me. I will do that for the bigger PRs :)

PR #783, opened 2020-10-22
Initial left & right skill select panel implementation. HeroSkill ser…

presiyan-ivanov:
…ialization cleanup.

- Clicking the active left/right skill now opens a skill select panel.
Only the available skills for the hero, which are valid for the panel type are shown.
Clicking on a skill from the skill select panel makes it the new active skill for the hero.

- Hovering a skill in the skill select panel shows the skill name +
skill description.

- New command which learns all skills for a specific
class(not persisted to a save file yet) - e.g. `learnskills ama` will learn
skills for the Amazon class.

- Initialize HeroSkill.shallowHeroSkill struct in the hero state factory, so we can use it
when we serialize the HeroSkill to packets/game save files.

- The parsed Skill.ListRow is now a number instead  of string.

Screenshot of the skill select panel:

od2_skills

### DIFF
--- a/d2common/d2enum/skill_class.go
+++ b/d2common/d2enum/skill_class.go
@@ -1,0 +1,73 @@
+package d2enum
+
+import "log"
+
+type SkillClass int
+
+const (
+	// SkillClassGeneric is ""
+	SkillClassGeneric SkillClass = iota
+	SkillClassBarbarian
+	SkillClassNecromancer
+	SkillClassPaladin
+	SkillClassAssassin
+	SkillClassSorceress
+	SkillClassAmazon
+	SkillClassDruid
+)
+
+// FromToken returns the enum which corresponds to the given class token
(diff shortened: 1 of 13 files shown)


gravestench (approved):
Looks good

gravestench:
the hover tooltip needs a semi-transparent box behind the text:
image

presiyan-ivanov:
> the hover tooltip needs a semi-transparent box behind the text:
> image

Yup, to keep the PR small I left that out from the initial implementation(I left a TODO   [here](https://github.com/OpenDiablo2/OpenDiablo2/pull/783/commits/82e015bcca52eaa77a672480a261e1f371700b13#diff-a9af951de50efb28911b3376f4a428db474daa1612f17f990aa7b0aec01696c8R43)  ). 

Also, hovering the active left/right skill should also display the tooltip, like it is for the popup panel(needs to be a cached image with text + background)

PR #808, opened 2020-10-25
Add close buttons to all panels

juander-ux:
Hi,

this is a small cleanup after my skilltree panel PR. It just adds the close buttons to all panels. While at it I fixed two minor linter cleanup mistakes. Is it possible to run the linter automatically on PR, so that we catch the problems before they get merged?

Cheers

### DIFF
--- a/d2game/d2player/game_controls.go
+++ b/d2game/d2player/game_controls.go
@@ -221,12 +221,12 @@ const (
 	menuLeftRectX,
 	menuLeftRectY,
 	menuLeftRectW,
-	menuLeftRectH = 0, 550, 800, 50
+	menuLeftRectH = 0, 0, 400, 600
 
 	menuRightRectX,
 	menuRightRectY,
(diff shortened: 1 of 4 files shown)


essial:
Once all lint errors are removed we will enforce linting on all PRs. When we started linting we had thousands of issues so there was no way to enforce it. I think thanks to grav we are getting close!

ThomasChr:
If wanted I can try to get a CI working. But that‘s something @essial need to do.

ThomasChr:
@essial 
Didn‘t read your message before posting mine. Yes, ci would be great!

juander-ux:
Are you using [golint](https://github.com/golang/lint)? I'm having a bit of trouble to run a locally from CLI.

gravestench:
@juander-ux get `golangci-lint`

Once installed, you can run from repo root
```
golangci-lint run ./...
```

PR #858, opened 2020-10-27
This CI should automatically build Linux and Windows Release Files with the Tag 'latest'

ThomasChr:
I did not manage to generate a MacOS Build
```
GOOS=darwin GOARCH=amd64 go build
```
does fail...

### DIFF
--- a/.github/workflows/pushToMaster.yml
+++ b/.github/workflows/pushToMaster.yml
@@ -37,6 +37,8 @@ jobs:
 
             -   name: Build binary
                 run: go build .
+
+            -   name: Upload binary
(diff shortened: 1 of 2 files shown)


ThomasChr:
To be clear: This is just a proposal with some example changes.
On my repo it worked and I would find it nice if there is a bleeding edge release available for people who just want to try it out without installing anything.
(Of course, they need to have the diablo MPQ files, but we can't change that)

gravestench:
what's wrong?

ThomasChr:
@essial Is working on a way to provide non-free ready built binaries.

PR #871, opened 2020-10-28
Dedicated server

Zaprit:
I wrote some basic dedicated server support for OpenDiablo2, plus if you control c it closes nicely


gravestench (changes requested):
I don't think that main is the right place for this to be added. Ideally, everything should be tucked into d2app, and main should just be calling d2app init stuff. Can you try to put the dedicated server initialization logic inside of d2app?

essial:
Your PR has merge conflicts after merging the ebiten v3 changes.

d2networking/dedicated_server.go
```diff
@@ -0,0 +1,74 @@
...
+	// messages later, sue me.
+	go func() {
+		for {
+			msgIn := <-in
+			/* For those who do not know an AND operation denoted by & discards bits which do not line up so for instance:
+			01011001 & 00000001 = 00000001 or 1
+			00100101 & 00000010 = 00000000 or 0
+			01100110 & 01100000 = 01100000 or 96
+			these can be used to have multiple messages in just 8 bits, that's a quarter of a rune in go!
+			*/
```
gravestench, line 61: This comment is not necessary. Your previous comment about using a bitmask is enough for those who are unfamiliar to google these details.

PR #872, opened 2020-10-28
Tooltip refactor

juander-ux:
Hi,

this PR refactors tooltips into their own ui element. Before skill_selection_panel and inventory were creating their own version, replicating a lot of code.

cheers

### DIFF
--- a/d2core/d2ui/tooltip.go
+++ b/d2core/d2ui/tooltip.go
@@ -1,0 +1,163 @@
+package d2ui
+
+import (
+	"strings"
+
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2interface"
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2util"
+	"github.com/OpenDiablo2/OpenDiablo2/d2core/d2gui"
+)
+
+const (
+	blackAlpha70 = 0x000000C8
+	screenWidth  = 800
+	screenHeight = 600
+)
+
+type Tooltip struct {
+	manager         *UIManager
+	lines           []string
+	label           *Label
+	backgroundColor int
+	x, y            int
+	originX         TooltipXOrigin
+	originY         TooltipYOrigin
+}
+
+type TooltipXOrigin = int
+type TooltipYOrigin = int
+const (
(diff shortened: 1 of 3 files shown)


gravestench:
We no longer accept PR's that introduce any lint errors. you should install `golangci-lint` and check your branch by running `golangci-lint run ./...` in the repo root.

juander-ux:
Huh? I guess my golangci-lint setup is broken then :)

gravestench:
If you're using goland, you need to add it as a file watcher after installing the plugin (File >> Settings >> Tools >> File Watchers)

juander-ux:
Thanks. I figured it out. Now I get the linter warnings. Though, I disagree with the linter on one statement
```
d2core/d2ui/tooltip.go:140:23: mnd: Magic number: 2, in <operation> detected (gomnd)
        halfW, halfH := maxW/2, maxH/2
                             ^
```
I disagree that this is a magic number. I don't see any benefit in replacing the 2 with a const. If you insist on the linter having the final say, fine. I'm just thinking that this is not a good style, as it just creates noise where the const is defined for no real benefit.

juander-ux:
Hmm, golangci-lint 1.27.0 and 1.31.0 seem to disagree :). I should install 1.27.0 on my machine as well.

gravestench:
If it's an integer you can shift right by 1 to divide by two, or you can specify a nolint directive. I agree, don't need a const for this.

Nolint directive would look like
```golang
// nolint:gomnd // reason for nolint
```

juander-ux:
> If it's an integer you can shift right by 1 to divide by two
I assumed the compiler is clever enough to figure this out. I like clarity more than cleverness :)

> // nolint:gomnd // reason for nolint
Ah cool, let me fix my PR :)

gravestench:
I mean, the `2` here _is a constant_, it wouldn't be the worst thing to actually make it a const within that scope, even if it is a super small gain

gravestench:
Either way, good work, and thank you for working on this. :+1:

essial:
If you wanna get _really_ specific, / 2 doesn't mean much on its own. That being said, silencing this lint error on this line is just fine by me.

Issue #799, opened 2020-10-25
Left/Right skill buttons need selected skills from character save file


gravestench:
in `GameControls.Load`, the character save file data should be used to select the currently selected skills for left/right mouse buttons.

gravestench:
I wonder if i can just assign @ThomasChr if i mention him in this issue...

ThomasChr:
Now I‘m not doing it ;-)

ThomasChr:
Can be closed!

Issue #867, opened 2020-10-28
Save inventory/equipment data to character save file


gravestench:
The items that need to be saved are:

- inventory
- equipped items
- cube
- stash
- mercenary
- corpse
- "cursor buffer"

The cursor buffer is whatever is currently held. This is essential, just think of getting disconnected while moving inventory!

currently blocking  #791 and #795

Zaprit:
i can do that

ThomasChr:
@Zaprit Any progress? Do you need help? Saving and loading for stats is implemented. So adding a first version for inventory shouldn‘t be a big deal anymore.

Zaprit:
@ThomasChr Sorry I completely forgot about this for a while, I'll get right on it.

gravestench:
@Zaprit Status?

Issue #815, opened 2020-10-25
Replace static health/mana/experience values with values from character save file


gravestench:
in `d2core/d2hero/hero_stats_state.go`

There are currently static test values that set the current helth, mana, and experience. These need to be removed and populated with values from the character save file.

Ganitzsh:
Is the Level also missing?

gravestench:
I think so, yes

ThomasChr:
I think this one can be close after the PR from @stephenhoran

gravestench:
resolved by #886

PR #883, opened 2020-10-29
Saving and loading game data, Fixes #868 #799

### LINKED ISSUE #868: Save currently selected skills in character save file
The skills (or skill indices, whatever) that are assigned to left/right mouse buttons need to be saved to the character save file.

There are 2 sets of left/right indices that need to be saved, one for main and another for secondary weapon+shield equipment slots.

currently blocking #799

ThomasChr:
It would be great if someone could take a quick look at the mess I made. It's far from being ready - but am I on the right track?

### DIFF
--- a/d2core/d2hero/hero_state.go
+++ b/d2core/d2hero/hero_state.go
@@ -7,14 +7,16 @@ import (
 
 // HeroState stores the state of the player
 type HeroState struct {
-	HeroName  string                         `json:"heroName"`
-	HeroType  d2enum.Hero                    `json:"heroType"`
-	HeroLevel int                            `json:"heroLevel"`
-	Act       int                            `json:"act"`
-	FilePath  string                         `json:"-"`
-	Equipment d2inventory.CharacterEquipment `json:"equipment"`
-	Stats     *HeroStatsState                `json:"stats"`
-	Skills    map[int]*HeroSkill             `json:"skills"`
-	X         float64                        `json:"x"`
-	Y         float64                        `json:"y"`
+	HeroName   string                         `json:"heroName"`
+	HeroType   d2enum.Hero                    `json:"heroType"`
+	HeroLevel  int                            `json:"heroLevel"`
+	Act        int                            `json:"act"`
+	FilePath   string                         `json:"-"`
+	Equipment  d2inventory.CharacterEquipment `json:"equipment"`
+	Stats      *HeroStatsState                `json:"stats"`
+	Skills     map[int]*HeroSkill             `json:"skills"`
+	X          float64                        `json:"x"`
+	Y          float64                        `json:"y"`
+	LeftSkill  *HeroSkill                     `json:"leftSkill"`
+	RightSkill *HeroSkill                     `json:"rightSkill"`
 }
--- a/d2core/d2hero/hero_state.go
+++ b/d2core/d2hero/hero_state.go
@@ -17,6 +17,6 @@ type HeroState struct {
 	Skills     map[int]*HeroSkill             `json:"skills"`
 	X          float64                        `json:"x"`
 	Y          float64                        `json:"y"`
-	LeftSkill  *HeroSkill                     `json:"leftSkill"`
-	RightSkill *HeroSkill                     `json:"rightSkill"`
+	LeftSkill  int                            `json:"leftSkill"`
+	RightSkill int                            `json:"rightSkill"`
 }
--- a/d2game/d2gamescreen/game.go
+++ b/d2game/d2gamescreen/game.go
@@ -313,8 +313,8 @@ func (v *Game) OnPlayerMove(targetX, targetY float64) {
 	worldPosition := v.localPlayer.Position.World()
 
 	playerID, worldX, worldY := v.gameClient.PlayerID, worldPosition.X(), worldPosition.Y()
-	createPlayerPacket := d2netpacket.CreateMovePlayerPacket(playerID, worldX, worldY, targetX, targetY)
-	err := v.gameClient.SendPacketToServer(createPlayerPacket)
+	createMovePlayerPacket := d2netpacket.CreateMovePlayerPacket(playerID, worldX, worldY, targetX, targetY)
+	err := v.gameClient.SendPacketToServer(createMovePlayerPacket)
 
 	if err != nil {
 		fmt.Printf(moveErrStr, v.gameClient.PlayerID, targetX, targetY)
--- a/d2game/d2gamescreen/game.go
+++ b/d2game/d2gamescreen/game.go
@@ -186,6 +186,10 @@ func (v *Game) OnUnload() error {
 		return err
 	}
 
+	if err := v.OnPlayerSave(); err != nil {
+		return err
+	}
+
 	if err := v.gameClient.Close(); err != nil {
 		return err
 	}
@@ -315,6 +319,18 @@ func (v *Game) OnPlayerMove(targetX, targetY float64) {
 	if err != nil {
 		fmt.Printf(moveErrStr, v.gameClient.PlayerID, targetX, targetY)
 	}
+}
+
+// OnPlayerSave instructs the server to save our player data
+func (v *Game) OnPlayerSave() error {
+	playerState := v.gameClient.Players[v.gameClient.PlayerID]
+	err := v.gameClient.SendPacketToServer(d2netpacket.CreateSavePlayerPacket(playerState))
+
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // OnPlayerCast sends the casting skill action to the server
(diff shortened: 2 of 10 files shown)


d2game/d2gamescreen/game.go
```diff
@@ -92,7 +92,7 @@ func CreateGame(
 		ticksSinceLevelCheck: 0,
 		mapRenderer: d2maprenderer.CreateMapRenderer(asset, renderer,
 			gameClient.MapEngine, term, startX, startY),
-		escapeMenu:    d2player.NewEscapeMenu(navigator, renderer, audioProvider, guiManager, asset),
+		escapeMenu:    d2player.NewEscapeMenu(navigator, renderer, audioProvider, guiManager, asset, gameClient),
```
juander-ux, line 95: Why are you giving the gameClient to the escape menu? You could also get away with just giving it the player, as this is all that you need.

d2game/d2player/escape_menu.go
```diff
@@ -414,6 +427,23 @@ func (m *EscapeMenu) showLayout(id layoutID) {
...
+			FilePath:   "",
+			Equipment:  *m.gameClient.Players[m.gameClient.PlayerID].Equipment,
+			Stats:      m.gameClient.Players[m.gameClient.PlayerID].Stats,
+			Skills:     m.gameClient.Players[m.gameClient.PlayerID].Skills,
+			X:          m.gameClient.Players[m.gameClient.PlayerID].Position.X(),
+			Y:          m.gameClient.Players[m.gameClient.PlayerID].Position.Y(),
+			LeftSkill:  m.gameClient.Players[m.gameClient.PlayerID].LeftSkill,
+			RightSkill: m.gameClient.Players[m.gameClient.PlayerID].RightSkill,
+		}
+		m.heroState.Save(&heroState)
```
juander-ux, line 445: I would not expect that the saving of the hero happens in the escape menu. I would rather put that into the player class as you only used properties from that class.  Then you can call that method here.

juander-ux (changes requested):

d2game/d2gamescreen/game.go
ThomasChr: Hey @juander-ux!
Thanks for taking a look at it.
Yes, only giving the player class would be a good idea. I will change that.

d2game/d2player/escape_menu.go
ThomasChr: Of this I was aware, and yes, that will be changed too. I already told grave in Discord that I don't find it very good that the esacpe menu saves a character. That doesn't make much sense.
After I got it to work I will also refactor that.

With this WIP: PR I only wanted to check if I'm completely off track here. But it does not seem so. Thanks for your help!

ThomasChr (commented):
Pr is WIP. Both problems will be changed in final PR

juander-ux:
I think you are on the right track with these changes.

d2game/d2player/escape_menu.go
```diff
@@ -427,22 +418,10 @@ func (m *EscapeMenu) showLayout(id layoutID) {
...
-			Equipment:  *m.gameClient.Players[m.gameClient.PlayerID].Equipment,
-			Stats:      m.gameClient.Players[m.gameClient.PlayerID].Stats,
-			Skills:     m.gameClient.Players[m.gameClient.PlayerID].Skills,
-			X:          m.gameClient.Players[m.gameClient.PlayerID].Position.X(),
-			Y:          m.gameClient.Players[m.gameClient.PlayerID].Position.Y(),
-			LeftSkill:  m.gameClient.Players[m.gameClient.PlayerID].LeftSkill,
-			RightSkill: m.gameClient.Players[m.gameClient.PlayerID].RightSkill,
+		err := m.player.Save()
+		if err != nil {
+			return
```
juander-ux, line 423: Minor nitpick. Can you print the error here, instead of silently failing?

juander-ux (approved):
Looks good to me. While reviewing, I did wonder why HeroState is no part of the player. Is this an oversight?

d2game/d2player/escape_menu.go
ThomasChr: I love nitpicks. Yes, printing it is!

ThomasChr:
HeroState seems to be a structure which is on file and used by the server. So it has nothing to do with the player directly. At least that's my conclusion.

ThomasChr:
It's untested at the moment and my linter flags not the same errors. So it's a little bit tricky. Just in case anyone is wondering why I'm pushing the hell out of it!

ThomasChr:
The client has nothing to do with loading and saving Hero's (for example it does not know the path of the used save file). So the server has to do that. I'll change my code.

ThomasChr:
Status: 
- [x] Saving Left and Right Skill to File (saving is done on (local) Serverside
- [x] Loading Left and Right Skill from File

I'm not quite sure how that is supposed to work with a remote Server. Maybe the remote Server sends it's client a SavePlayer Packet? Or Maybe (in case it's like battlenet) the Character is saved on the server side.
Anyone an idea?

d2networking/d2netpacket/packet_save_player.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/OpenDiablo2/OpenDiablo2/d2core/d2map/d2mapentity"
+	"github.com/OpenDiablo2/OpenDiablo2/d2networking/d2netpacket/d2netpackettype"
+)
+
+// SavePlayerPacket has the actual selected left and right skill
+// the Server has to check if these skills are actually allowed for the Player
+type SavePlayerPacket struct {
+	LeftSkill  int `json:"leftSkill"`
+	RightSkill int `json:"rightSkill"`
+}
```
gravestench, line 16: We don't save any other data?

gravestench (approved):

d2networking/d2netpacket/packet_save_player.go
ThomasChr: The server already has got all of the base hero data. So we only need to send all of the things he can‘t possible know and need to write them back in its herodata Struct. There is much more we need to save, for example equipment data, but theres an extra issue for that #867 
So the structure will surely grow bigger.

We also need to think about when we send the Server our changed data and when it should save it to a file (or only update it‘s herodata without saving it). Surely on connection close (like in this code) that‘s a good idea to do. But maybe we need to send those packets more often, like when the equipment changes. The server not really does need to know which left and right skill is active on my client (it only needs to know if I have a skill when casting) but maybe it need to get notified everytime I change my equipment... I‘m not quite sure about this.
That sad, I think this code is a step into the right direction.
Thanks for your thoughts about this.

ThomasChr:
I think this code could also be helpfull for #815

d2core/d2hero/hero_state.go
```diff
@@ -7,14 +7,16 @@ import (
...
+	HeroType   d2enum.Hero                    `json:"heroType"`
+	HeroLevel  int                            `json:"heroLevel"`
+	Act        int                            `json:"act"`
+	FilePath   string                         `json:"-"`
+	Equipment  d2inventory.CharacterEquipment `json:"equipment"`
+	Stats      *HeroStatsState                `json:"stats"`
+	Skills     map[int]*HeroSkill             `json:"skills"`
+	X          float64                        `json:"x"`
+	Y          float64                        `json:"y"`
+	LeftSkill  int                            `json:"leftSkill"`
```
presiyan-ivanov, line 20: Could be named LeftSkillID and RightSkillID, although I think its better to change the type to *HeroSkill, and re-use what was previously done for serializing/deserializing those, as it is the same use-case.

ThomasChr: Thanks @presiyan-ivanov for taking a look! I thought about both of this and I saw the serialization function for the skill, also my forst implementation used exactly that. But I saved a sub json object in the json with the skillPoints. I didn‘t like it because the skillPoints are not needed at this point and so I only used the skillID which made it also a little bit clearer when loading the Skills.
Also I named them LeftSkill and RightSkill because thats what they are called in the PlayerStruct (but yes, they are of type heroSkill, not only a skillID). My heart does not rely on both of these decisions and if you really want, I could change it. Up to you!

presiyan-ivanov: Good point about the skillPoints being redundant in this scenario. Then maybe just renaming to LeftSkillID & RightSkillID might be best in this case

ThomasChr: @presiyan-ivanov 
Yes, naming them LeftSkillID & RightSkillID
But now it's merged. Do you think that's worth a new PR or should I just hope to think about it and change it in my next PR?
Also you can change it if you do a PR in the next days...

presiyan-ivanov: I think its overkill to open a PR just for that, you can do it with another PR, and it is not really that big of a deal if it stays like this anyway :)

PR #888, opened 2020-10-31
removed global variable for singleton game server

### LINKED ISSUE #824: Remove global variable for singleton game server
We need to remove the singleton server global variable, 'nuff said.

gurkankaymak:
fixes #824


essial:
Does spawn item and cast still work? I noticed those removed from the switch.

gurkankaymak:
not specifically tested for them but they are not removed, refactored to be in the same case since inside of the case were the same

PR #893, opened 2020-11-01
Feat(KeyMapping): Adds a configurable keymap to GameControls, resolves #793

### LINKED ISSUE #793: Need an abstraction between input events and game engine events
Currently, all interaction in the game engine is hard-coded for specific input events (keyboard keys, mouse buttons).

However, we need some kind of abstraction between input events and the game events they are mapped to. 

We also need this mapping of input to game events to be configurable via the `config.json` file.

Ganitzsh:
+ Updates help overlay to use it
#793

### DIFF
--- a/d2common/d2enum/game_event.go
+++ b/d2common/d2enum/game_event.go
@@ -1,0 +1,92 @@
+package d2enum
+
+type GameEvent int
+
+const (
+	// ToggleGameMenu will display the game menu
+	ToggleGameMenu GameEvent = iota + 1
+
+	// ToggleCharacterPanel toggles the player's character sheet
+	ToggleCharacterPanel
+	// ToggleInventoryPanel toggles the player's inventory
+	ToggleInventoryPanel
+	// TogglePartyPanel toggles the party manager
+	TogglePartyPanel
+	// ToggleSkillTreePanel toggles the skill tree
+	ToggleSkillTreePanel
+	// ToggleHirelingPanel toggles the hirelings manager
+	ToggleHirelingPanel
+	// ToggleQuestLog toggles the quest log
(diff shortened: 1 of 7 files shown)


d2common/d2enum/game_event.go
```diff
@@ -0,0 +1,118 @@
...
+type GameEvent int
+
+const (
+	// ToggleGameMenu will display the game menu
+	ToggleGameMenu GameEvent = iota + 1
+
+	// ToggleCharacterPanel toggles the player's character sheet
+	ToggleCharacterPanel
+	// ToggleInventoryPanel toggles the player's inventory
+	ToggleInventoryPanel
```
gravestench, line 13: to appease the linter, you could have put `// Game events` above the const declaration line.

d2game/d2player/help_overlay.go
```diff
@@ -1,11 +1,11 @@
-// Package help contains the in-game diablo2 help panel
-package help
+package d2player
```
gravestench, line 2: I think what @brendanporter did with isolating individual panels to another sub-package is a smart idea; it keeps `d2player/` from becoming monolithic. I'm not sure about moving the help overlay back into `d2player/`.

@Ganitzsh Was this just to keep everything consistent?

@essial How do you feel about this?

gravestench (approved):
If we all agree that we want to make each panel in `d2player` into a sub-package, I can go ahead and make an issue for it.

As it stands, I think that it's ok to merge. We can decide what we want to at a later time.

Good work @Ganitzsh !

gravestench:
@Ganitzsh can you put your name into the `CONTRIBUTORS` file?

PR #903, opened 2020-11-05
Hotkeys in Main Menu

gucio321:
In this PR I implemented Escape button usage in game main menu.


gravestench:
@gucio321 We don't allow any PR's to merge if they contain lint errors. You should install `golangci-lint` and run it from the repo directory to see the errors:
```bash
golangci-lint run ./...
```

PR #911, opened 2020-11-09
d2ui refactor

juander-ux:
Hi,

this is the first step in removing the render function from game_control. For now I only reworked the skilltree panel, but other panel will follow soon. Since this has many invasive changes, I'd rather merge it now, instead of me and others having annoying rebasing work. Features of this PR:

* All ui elements of d2ui now implement the Widget interface
* All ui elements of d2ui now use a BaseWidget that has the shared functionality, like Set/GetVisible
* Introduce WidgetGroups, which allow us to group widgets together to do bulk operations on the Widgets. For now this is only SetVisible
* Introduce CustomWidget, which can be used give a custom render function to a widget.
* Introduce RenderPriorities for WidgetGroups/Widgets which determines the order in which widgets are rendered

Since the why is important in this PR, **please do not squash it into one commit**.

Cheers,

### DIFF
--- a/d2core/d2ui/button.go
+++ b/d2core/d2ui/button.go
@@ -242,32 +242,28 @@ var _ Widget = &Button{} // static check to ensure button implements widget
 
 // Button defines a standard wide UI button
 type Button struct {
-	manager               *UIManager
+	*BaseWidget
 	buttonLayout          ButtonLayout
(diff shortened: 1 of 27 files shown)


gravestench (approved):
Good work.

PR #912, opened 2020-11-10
Cinematics menu

gucio321:
Hi ther,
In this pull request I've added cinematics menu.
Additionally, I've fixed bug with black background in Server IP menu.


gravestench:
Good catch on that tcp/up screen bug!

PR #914, opened 2020-11-11
d2ui/drawable refactor

juander-ux:
Hi,

as suggested in Discord by @gravestench, I remove the return of error in the Render() method. While at that I reverted my hacks, that introduced separate RenderNoError() methods, as this was a mistake.

cheers,


essial:
Great! Make sure you add yourself to the contributors file if you haven't already so you show up in the credits!

juander-ux:
Hi @essial, being in the credits is not important to me. 
cheers,

PR #916, opened 2020-11-11
added user IP in TCPIP menu

gucio321:
Hi ther,
So again, now without fatal code conflicts ;-)
In this PR, I'm adding label with IP address in TCPIP menu. I've also a bit corrected Join Game and Host Game buttons position.

### DIFF
--- a/d2game/d2gamescreen/main_menu.go
+++ b/d2game/d2gamescreen/main_menu.go
@@ -270,7 +270,7 @@ func (v *MainMenu) createLabels(loading d2screen.LoadingState) {
 
 	v.machineIP = v.uiManager.NewLabel(d2resource.Font24, d2resource.PaletteUnits)
 	v.machineIP.Alignment = d2gui.HorizontalAlignCenter
-	v.machineIP.SetText("Your IP adress is:\n" + v.getLocalIP())
+	v.machineIP.SetText("Your IP address is:\n" + v.getLocalIP())
 	v.machineIP.Color[0] = rgbaColor(lightYellow)
 	v.machineIP.SetPosition(machineIPX, machineIPY)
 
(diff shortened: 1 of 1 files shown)


d2game/d2gamescreen/main_menu.go
```diff
@@ -612,3 +620,25 @@ func (v *MainMenu) onBtnTCPIPCancelClicked() {
 func (v *MainMenu) onBtnTCPIPOkClicked() {
 	v.navigator.ToCharacterSelect(d2clientconnectiontype.LANClient, v.tcpJoinGameEntry.GetText())
 }
+
+// getLocalIP returns local machine IP address
+func (v *MainMenu) getLocalIP() string {
+	conn, err := net.Dial("udp", "8.8.8.8:80")
+
```
gravestench, line 627: This is assuming that you're on the internet. What happens when you're just on a private LAN...?

gravestench (approved):

d2game/d2gamescreen/main_menu.go
gravestench: also, `8.8.8.8:80` is magic; an assumption that the significance of this value is implicit. In 20 years, what will someone looking at this code think? just my 2 cents.

essial: What about `LookupHost("localhost")` or something?

gucio321: Yah, I also think, that's not a best solution, but I'm not enough in go to do it correct. If you have some time, correct it please or simply I'll remove this function and replace it with TODO

gravestench: This should be good for now. I'll make an issue for it.

PR #925, opened 2020-11-14
implemented logger in d2gamescreen

gucio321:
Hi there,
no more import "log" in module d2gamescreen


d2app/app.go
```diff
@@ -829,7 +829,7 @@ func (a *App) quitGame() {
 }
 
 func (a *App) enterGuiPlayground() {
-	a.screen.SetNextScreen(d2gamescreen.CreateGuiTestMain(a.renderer, a.guiManager, a.asset))
+	a.screen.SetNextScreen(d2gamescreen.CreateGuiTestMain(a.renderer, a.guiManager, d2util.LogLevelDefault, a.asset))
 }
```
gravestench, line 834: we should use the log level that is defined in the config

gravestench (changes requested):
Dont use the default log level, pass the `LogLevel` in from `App.Config.LogLevel` when you create the screens.

d2app/app.go
gucio321: ok @gravestanch I've corrected it.

gravestench (approved):

gucio321:
I've added logger to d2game/d2player
sory for lints, I'll fix it tomorrow :-)

gucio321:
Ok, I think, this is complete

d2game/d2player/globeWidget.go
```diff
@@ -95,28 +88,43 @@ func newGlobeWidget(ui *d2ui.UIManager, x, y int, gtype globeType, value, valueM
...
 		value:      value,
 		valueMax:   valueMax,
 		globe:      globe,
 		overlap:    overlap,
 	}
+
+	gw.logger = d2util.NewLogger()
+	gw.logger.SetLevel(l)
+	gw.logger.SetPrefix(logPrefix)
+
```
gravestench, line 102: I think that this is problematic. Do individual widgets need logger instances?

should we not instead return these errors to the HUD and let the HUD report errors?

d2game/d2player/skillicon.go
```diff
@@ -41,11 +37,24 @@ func newSkillIcon(ui *d2ui.UIManager, baseSprite *d2ui.Sprite, skill *d2hero.Her
...
 }
 
+type skillIcon struct {
+	*d2ui.BaseWidget
+	lvlLabel *d2ui.Label
+	sprite   *d2ui.Sprite
+	skill    *d2hero.HeroSkill
+
+	logger *d2util.Logger
+}
```
gravestench, line 56: another widget that has it's own logger. maybe this is not such an issue if you assign to it an existing logger (like from the HUD or some menu)

gravestench (commented):
I don't think that widgets should create loggers, they should likely use a parent logger instance (like the UI Manager logger instance). 

@essial thoughts?

d2game/d2player/globeWidget.go
gucio321: yes, we can return errors to HUD, but ONLY ERRORS. What about logger.Info like in help_overlay:go:202?

d2game/d2player/skillicon.go
gucio321: maybe, so could you give some example. what should I type instead of `logger *d2util.Logger`?

gravestench:
After a lot of discussion with @gucio321, it's not so bad if widgets have loggers. An alternate approach of using the logger of whatever was using the widget was considered, but was more trouble than it was worth. If there are no objections I would like to merge.

PR #933, opened 2020-11-18
save player: level, health, mana, experience

### LINKED ISSUE #865: Save current health, mana, level, and experience to the character save file.
blocking #815

ThomasChr:
Fixes #865
Draft because: Not tested yet.


ThomasChr:
We're now saving the complete Stats Struct from the Client - and currently selected Left/Right-Skill.
It's easy to change that anytime if we don't trust the client with some of the fields.

gravestench (approved):

PR #937, opened 2020-11-20
Getting your local IP should not depend on an Internet connection

### LINKED ISSUE #917: TCP/IP screen should not depend on an internet connection
Related to PR #916 

The IP address displayed in the `Join Game` main menu should not be dependent on an active internet connection.

Currently, the IP address is resolved by connecting to google's dns server at `8.8.8.8:80`, but this doesn't account for LAN players.

The offending code is in `d2game/d2gamescreen/main_menu.go`

ThomasChr:
Fixes #917

I copied nearly all of the code from here: https://stackoverflow.com/a/28862477
(the code I deleted was also very similar to code from this Thread, btw...)

Should we mark that anywhere?
I have no idea how that could affect our code if we copy a part from stackoverflow.
And if there are no legal issues (I tried to find an answer for that and it doesn't seem to be a problem), shouldn't we note the original author somewhere?

### DIFF
--- a/d2game/d2gamescreen/main_menu.go
+++ b/d2game/d2gamescreen/main_menu.go
@@ -633,6 +633,7 @@ func (v *MainMenu) onBtnTCPIPOkClicked() {
 
 // getLocalIP returns local machine IP address
 func (v *MainMenu) getLocalIP() string {
+	// https://stackoverflow.com/a/28862477
 	host, _ := os.Hostname()
 	addrs, _ := net.LookupIP(host)
 
--- a/d2game/d2gamescreen/main_menu.go
+++ b/d2game/d2gamescreen/main_menu.go
@@ -633,20 +633,16 @@ func (v *MainMenu) onBtnTCPIPOkClicked() {
 
 // getLocalIP returns local machine IP address
 func (v *MainMenu) getLocalIP() string {
-	conn, err := net.Dial("udp", "8.8.8.8:80")
+	host, _ := os.Hostname()
+	addrs, _ := net.LookupIP(host)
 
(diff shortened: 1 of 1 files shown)


ThomasChr:
And YES, there is still kind of a problem with the code:
You're getting the IP (v4) Address from ONE of your network interfaces. There could also be IPv6 Addresses and more IPv4 Addresses. You have no clue which address has internet access and which does not, and which would be the one the Player should use. So the display is helpful if you only got one interface and using Ipv4.

d2game/d2gamescreen/main_menu.go
```diff
@@ -633,20 +633,16 @@ func (v *MainMenu) onBtnTCPIPOkClicked() {
...
-		v.logger.Error(err.Error())
-		return "unexpected error occurred while closing test connection"
+	for _, addr := range addrs {
+		if ipv4 := addr.To4(); ipv4 != nil {
+			return ipv4.String()
+		}
 	}
 
-	localAddr := conn.LocalAddr().(*net.UDPAddr)
+	v.logger.Error("no IPv4 Address could be found")
```
gucio321, line 645: I think, It should be LogLevelWarning

ThomasChr: Yeah. Why not. It's not a really bad error. But of course if you can't find any local interface than surely you won't be able to play over Network I guess.
I'll change it.

gravestench (approved):
Looks good. If you want, you could link to the stack overflow post in a comment.

ThomasChr:
Added Stackoverflow Link. Thanks @gravestench

PR #941, opened 2020-11-21
close button label in help overlay

gucio321:
Hi there,
I've noticed some minor bug (it isn't visible in english locale, but using another language, w can see it):
close button label in help overlay is alignet to left, so i another locale, where "cloes" word has another lenght it is alignet to left side of close button, so...


gravestench:
A picture is worth a thousand words... :)

gravestench (approved):

PR #942, opened 2020-11-21
save act and difficulty. Fixes #866

### LINKED ISSUE #866: Save last known difficulty and location to character save file
The location is just a reference to an Act, this is used to spawn the player in the correct town when joining a game.

In retail d2, when changing difficulties, the player is always spawned in act 1.

currently blocking #829

ThomasChr:
I'm not quite sure if that's all or if I would like to change something...
So Draft for now until I had time to sleep and think about it :)
Fixes #866

### DIFF
--- a/d2core/d2map/d2mapentity/factory.go
+++ b/d2core/d2map/d2mapentity/factory.go
@@ -102,6 +102,7 @@ func (f *MapEntityFactory) NewPlayer(id, name string, x, y, direction int, heroT
 		isInTown:     true,
 		isRunning:    false,
 		Gold:         gold,
+		Act:          1,
 	}
 
 	result.mapEntity.uuid = id
--- a/d2core/d2records/difficultylevels_loader.go
+++ b/d2core/d2records/difficultylevels_loader.go
@@ -1,6 +1,7 @@
 package d2records
 
 import (
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2enum"
 	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2fileformats/d2txt"
 )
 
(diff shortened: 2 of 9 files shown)


gucio321:
Don't you think, `Difficulty` will be better as intager? In my opinion it'll be easyer to process it latter.
we can describe somewgere difficulty mods e.g.
```const (
   DifficultyUnknown = iota
   DifficultyNormal
   DifficultyNightmare
   DifficultyHell
)```

**Comment by ThomasChr** [comment]
@gucio321 
Yes! That was something I thought about. Using the code like you suggested would be much better then saving a string!
Thanks!

**Comment by gravestench** [comment]
just fyi, there is a txt file with records for the difficulty, and i think we already have an enum for difficulty in `d2common/d2enum/`

**Comment by ThomasChr** [comment]
@gravestench Thanks!

**Comment by ThomasChr** [comment]
Seems to work just fine and has been rebased on the most recent master!

File: d2core/d2map/d2mapentity/factory.go [hunk 1]
```diff
@@ -102,6 +102,7 @@ func (f *MapEntityFactory) NewPlayer(id, name string, x, y, direction int, heroT
 		isInTown:     true,
 		isRunning:    false,
 		Gold:         gold,
+		Act:          1,
```
gravestench, line 105: This would set new players to start in `DifficultyNightmare`

ThomasChr: Does it? It only sets the Act, which is an attribute of the player. The difficulty level is not an attribute of the player, it gets set nowhere at the moment and is only used in the save player function where it it hardcoded to d2enum.DifficultyNormal.

d2core/d2records/difficultylevels_loader.go
```diff
@@ -27,7 +28,14 @@ func difficultyLevelsLoader(r *RecordManager, d *d2txt.DataDictionary) error {
...
 		}
-		records[record.Name] = record
+		switch record.Name {
+		case "Normal":
+			records[d2enum.DifficultyNormal] = record
+		case "Nightmare":
+			records[d2enum.DifficultyNightmare] = record
+		case "Hell":
+			records[d2enum.DifficultyHell] = record
+		}
```
gravestench, line 38: Doesn't this couple the loader to expect explicitly the strings "Normal", "Nightmare" and "Hell" ? what happens when Johnny Modmaker comes along and adds an "Easy" difficulty?

ThomasChr: Yep. You're right. If we really want to allow any difficulty we can't use a difficulty enum, can we?

gravestench: For the meantime, i suppose it is okay, but i can see us moving all of the hardcoded data somewhere like an `OpenDiablo2.mpq` in the future.

PR #943, opened 2020-11-22
gold

gucio321:
Hi there,
in this PR I'm adding:
* gold value to save file
* gold button & label in inventory window
* types of "square" buttons in d2ui/button.go


gravestench (approved):
Good work!

PR #950, opened 2020-11-25
Hotfix: necromancer animation in select hero class menu

gucio321:
Hi there,
Next minor bug fixed: necromancer animation in select hero class menu covered up his tescription


gucio321:
sorry so many unnecessary commits attachet, I forgot to update this branch ;-)

Issue #944, opened 2020-11-23
differences betwen string tables


gucio321:
To make OD2 multi-language, labels in Main Menu must be related with string tables, but there are differences in names of string in this tables.
for example for Exit button label:
label "EXIT DIABLO II" in english table has name "#1625", but in polish table it is under the name "#1627"

gucio321:
Only one known and unsolved bug left:
https://github.com/OpenDiablo2/OpenDiablo2/blob/e982430c5533cb84530ca7542064bc1c7f124048/d2game/d2gamescreen/select_hero_class.go#L707

PR #951, opened 2020-11-25
Multi language labels

gucio321:
Hi there,
Labels in Main Menu should now depends on string tables.
I don't say, that's a perfect solution of #944, but it works for now.
Let me know, is main_menu.go right place to put this "table interpretation", or maybe better to put it into d2resource?

### DIFF
--- a/d2common/d2resource/languages_map.go
+++ b/d2common/d2resource/languages_map.go
@@ -50,6 +50,7 @@ func GetFontCharset(language string) string {
 	return charset[language]
 }
 
+// GetLabelModifier returns modifier for language
 /* modifiers for labels (used in string tables)
 modifier is something like that:
 english table:       polish table:
@@ -62,7 +63,6 @@ So, GetLabelModifier returns value of offset in locale languages table
 */
 // some of values need to be set up. For now values with "checked" comment
 // was tested and works fine in main menu.
-// GetLabelModifier returns modifier for language
 func GetLabelModifier(language string) int {
 	modifiers := map[string]int{
 		"ENG": 0, // (English) // checked
--- a/d2common/d2resource/languages_map.go
+++ b/d2common/d2resource/languages_map.go
@@ -50,7 +50,9 @@ func GetFontCharset(language string) string {
 	return charset[language]
 }
 
-// modificators for labels (used in string tables
+// modificators for labels (used in string tables)
+// some of values need to be set up. For now values with "checked" comment
+// was tested and works fine in main menu.
 func getModifiers() map[string]int {
 	return map[string]int{
 		"ENG": 0, // (English) // checked
--- a/d2common/d2resource/languages_map.go
+++ b/d2common/d2resource/languages_map.go
@@ -51,7 +51,7 @@ func GetFontCharset(language string) string {
 }
 
 // modificators for labels (used in string tables
-func getModificators() map[string]int {
+func getModifiers() map[string]int {
 	return map[string]int{
 		"ENG": 0, // (English) // checked
 		"ESP": 0, // (Spanish)
@@ -69,8 +69,8 @@ func getModificators() map[string]int {
 }
 
 // GetFontCharset returns modificator for language
-func GetLabelModificator(language string) int {
-	mod := getModificators()
+func GetLabelModifier(language string) int {
+	mod := getModifiers()
 
 	return mod[language]
 }
--- a/d2common/d2resource/languages_map.go
+++ b/d2common/d2resource/languages_map.go
@@ -68,7 +68,7 @@ func getModifiers() map[string]int {
 	}
 }
 
-// GetFontCharset returns modificator for language
+// GetLabelModifier returns modificator for language
 func GetLabelModifier(language string) int {
 	mod := getModifiers()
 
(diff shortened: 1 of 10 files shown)


d2game/d2gamescreen/main_menu.go
```diff
@@ -83,6 +83,142 @@ type BuildInfo struct {
...
+		1704, // nec description
+		1709, // barb description
+		1710, // sorc description
+		1711, // pal description
+
+		/* these items are for battle.net multiplayer
+		labels and are not used yet
+		in addition, as many elements as the value
+		of the highest modifier must be replaced*/
+		1712, // not used here
```
gravestench, line 199: Why put this here, then? is this value used?

d2common/d2resource/languages_map.go
```diff
@@ -49,3 +49,30 @@ func GetFontCharset(language string) string {
...
+		"RUS": 0, // (Russian) // checked
+	}
+}
+
+// GetLabelModifier returns modificator for language
+func GetLabelModifier(language string) int {
+	mod := getModifiers()
+
+	return mod[language]
+}
```
gravestench, line 78: replace all `modificator` with `modifier`

~~How are these integers used... ?~~ I think you should state that this table is used as an index offset for string keys of the form `#1234`

also, you could use just one function here:
```golang
// GetLabelModifier returns modificator for language
func GetLabelModifier(language string) int {
	modifiers := map[string]int{
		"ENG": 0, // (English) // checked
		"ESP": 0, // (Spanish)
		"DEU": 0, // (German) // checked
		"FRA": 0, // (French)
		"POR": 0, // (Portuguese)
		"ITA": 0, // (Italian)
		"JPN": 0, // (Japanese)
		"KOR": 0, // (Korean)
		"SIN": 0, //
		"CHI": 0, // (Chinese)
		"POL": 1, // (Polish) // checked
		"RUS": 0, // (Russian) // checked
	}

	return modifiers[language]
}
```

d2game/d2gamescreen/main_menu.go
gucio321: sorry, wrong comment her

d2game/d2gamescreen/main_menu.go
```diff
@@ -234,6 +374,10 @@ func (v *MainMenu) loadBackgroundSprites() {
 	v.serverIPBackground.SetPosition(serverIPbackgroundX, serverIPbackgroundY)
 }
 
+func translateLabel(label int, lng string, asset *d2asset.AssetManager) string {
+	return asset.TranslateString(fmt.Sprintf("#%d", baseLabelNumbers(label+d2resource.GetLabelModifier(lng))))
+}
+
```
gravestench, line 380: I don't think that this logic belongs in the main menu. Retrieving the correct string for the current locale should be something that the asset manager deals with, not the main menu.

d2game/d2gamescreen/main_menu.go
```diff
@@ -92,6 +162,7 @@ func CreateMainMenu(
 	audioProvider d2interface.AudioProvider,
 	ui *d2ui.UIManager,
 	buildInfo BuildInfo,
+	lng string,
```
gravestench, line 165: is this used anymore?

gucio321: no, I forgot to remove.

d2common/d2util/stringutils.go
```diff
@@ -135,11 +135,6 @@ func SplitIntoLinesWithMaxWidth(fullSentence string, maxChars int) []string {
 	return lines
 }
 
-// SplitIntoLinesWithMaxWidthOneLine does the same as SplitIntoLinesWithMaxWidth but return string with newline char
-func SplitIntoLinesWithMaxWidthOneLine(s string, l int) string {
-	return strings.Join(SplitIntoLinesWithMaxWidth(s, l), "\n")
-}
-
```
gravestench, line 142: Thank you fr removing this :)

d2game/d2gamescreen/main_menu.go
```diff
@@ -83,6 +84,76 @@ type BuildInfo struct {
...
+	hardCoreLabel
+	selectHeroClassLabel
+	amazonDescr
+	necromancerDescr
+	barbarianDescr
+	sorceressDescr
+	paladinDescr
+
+	//nolint:deadcode,varcheck,unused // will be used, or is need
+	notUsed1 // this position isn't used here.
```
gravestench, line 134: i think you can put `_` if you want to skip an index here

d2game/d2gamescreen/main_menu.go
```diff
@@ -83,6 +84,76 @@ type BuildInfo struct {
...
+	epilogueLabel
+	selectCinematicLabel
+
+	// multiplayer menu labels
+	//nolint:deadcode,varcheck,unused // will be used, or is need
+	openBattleNetLabeL
+	tcpIPGameLabel
+	tcpIPOptionsLabel
+	tcpIPHostGameLabel
+	tcpIPJoinGameLabel
```
gravestench, line 113: This is ok for now, but i do think that this enum belongs elsewhere.

gravestench (approved):

Issue #949, opened 2020-11-25
Minibar does not open if another window is open


ThomasChr:
**Describe the bug**
minibar does not open anymore if inventory or skill tree is open

**To Reproduce**
Open the skill tree and try to open the minibar at the bottom. It won't open.

**Expected behavior**
it should open

gucio321:
Yah, it is expected:
```        if !m.movedLeft && !m.movedRight {
                m.panelGroup.SetVisible(true)
        }
```

gucio321:
ok, fixed in #953

PR #953, opened 2020-11-26
Hotfix: minipanel bugs

gucio321:
Hi there,
minor bug: minipanels labels didn't get back to orginal position after closing char stats panel
also solves #949

### DIFF
--- a/d2game/d2player/mini_panel.go
+++ b/d2game/d2player/mini_panel.go
@@ -262,9 +262,7 @@ func (m *miniPanel) Toggle() {
 }
 
 func (m *miniPanel) Open() {
-	if !m.movedLeft && !m.movedRight {
-		m.panelGroup.SetVisible(true)
-	}
+	m.panelGroup.SetVisible(true)
 
 	if !m.menuButton.GetToggled() {
 		m.menuButton.Toggle()
--- a/d2game/d2player/mini_panel.go
+++ b/d2game/d2player/mini_panel.go
@@ -294,6 +294,7 @@ func (m *miniPanel) moveRight() {
 
 func (m *miniPanel) undoMoveRight() {
 	m.panelGroup.OffsetPosition(-panelOffsetRight, 0)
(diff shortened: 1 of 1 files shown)


juander-ux:
Good catch :). Looks good to me.

ThomasChr:
@gucio321 
If you want you could also take a look at #949

d2game/d2player/mini_panel.go
```diff
@@ -262,9 +262,7 @@ func (m *miniPanel) Toggle() {
 }
 
 func (m *miniPanel) Open() {
-	if !m.movedLeft && !m.movedRight {
-		m.panelGroup.SetVisible(true)
-	}
+	m.panelGroup.SetVisible(true)
```
juander-ux, line 265: I don't think this is the right solution. If the both the left and the right panel are open, we should not be able open the minipanel.

juander-ux (changes requested):

d2game/d2player/mini_panel.go
gucio321: thanks @juander-ux. Now should be OK

d2game/d2player/mini_panel.go
```diff
@@ -262,7 +262,7 @@ func (m *miniPanel) Toggle() {
 }
 
 func (m *miniPanel) Open() {
-	if !m.movedLeft && !m.movedRight {
+	if !(m.movedRight && m.movedLeft) {
```
juander-ux, line 265: Looks good! Thanks :)

juander-ux (approved):

d2game/d2player/mini_panel.go
gucio321: one bracket, but a big change :smile: Thanx for review @juander-ux

d2game/d2player/mini_panel.go
gravestench: From retail d2:
ezgif-3-971b7e263fe4

PR #954, opened 2020-11-27
colored credits

gucio321:
Hi there,
I've remade credits text, to use `d2ui.ColorTokenize`


ThomasChr:
You seem to be on a run. I like it!
Good Work!

PR #957, opened 2020-12-02
Hotfix: minor bugs in quest log

gucio321:
Hi there,
small bug: I've confused order of args in call to `newQuestLog`


gucio321:
it also fixes one more minor bug:
ther is no "KeyAlt" position in string tables. It should be "KeyMenu" We can see it on help screen ("Hold down <**KeyAlt**> to highlight items on the ground")

PR #962, opened 2020-12-07
Move Gold Panel

gucio321:
Hi there,
The major features added in this PR is "move Gold panel".
Over that, I've moved stuff responsible for AssetManager.TranslateLabel(...) (#951)  to d2enum, so OkButton's label is now displayed correctly (in another then english language).


d2core/d2ui/textbox.go
```diff
@@ -77,7 +86,21 @@ func (v *TextBox) OnKeyChars(event d2interface.KeyCharsEvent) bool {
...
 	if len(newText) > 0 {
-		v.text += newText
+		if v.isNumberOnly {
+			number, err := strconv.Atoi(v.text + newText)
+			if err != nil {
+				v.Debugf("Unable to convert string %s to intager: %s", v.text+newText, err)
+				return false
+			}
+
+			if number <= v.maxValue {
```
gravestench, line 96: shouldn't this be `number < v.maxValue` ?

gravestench, line 96: for this whole block of code, i would suggest inverting some of these if statements to reduce the nesting.
something like this:
```golang
func (v *TextBox) OnKeyChars(event d2interface.KeyCharsEvent) bool {
	if !v.isFocused || !v.visible || !v.enabled {
		return false
	}

	newText := string(event.Chars())
	if !(len(newText) > 0) {
		return false
	}
	
	if !v.isNumberOnly {
		v.text += newText
		return true
	}

	number, err := strconv.Atoi(v.text + newText)
	if err != nil {
		v.Debugf("Unable to convert string %s to intager: %s", v.text+newText, err)
		return false
	}

	if number <= v.maxValue {
		v.text += newText
	} else {
		v.text = strconv.Itoa(v.maxValue)
	}
	
	return true
}
```

gravestench: oh, also, `integer` instead of `intager` in the debug print.

d2game/d2player/game_controls.go
```diff
@@ -444,6 +444,12 @@ func (g *GameControls) onEscKey() {
...
 	if g.inventory.IsOpen() {
+		if g.inventory.moveGoldPanel.IsOpen() {
+			g.inventory.moveGoldPanel.Close()
+
+			return
+		}
+
 		g.inventory.Close()
 
 		escHandled = true
```
gravestench, line 455: This section of the code does not scale well at all, i think it needs a refactor. I'm thinking that as panels are opened/closed they should be pushed/popped from a stack `[]interface{}`. This could be stored inside of `GameControls` as `GameControls.panelStack` or something similar.

```golang
type panel interface {
	IsOpen() bool
	Close()
}

type panelContainer interface {
	HasOpenPanels() bool
	ClosePanels()
}

func (g *GameControls) onEscKey() {
	// special case for opening the escape menu when panel stack is empty
	if len(g.panelStack) == 0 {
		g.escapeMenu.Open()

		g.panelStack = append(g.panelStack, g.escapeMenu)

		g.updateLayout()

		return
	}

	// generic case, pop the last thing off the, close it, and then update
	for idx := range g.panelStack {
		escHandled := false
		
		if p, isPanel := g.panelStack[idx].(panel); isPanel {
			p.Close()
			escHandled = true
		}

		if p, isPanelContainer := g.panelStack[idx].(panelContainer); isPanelContainer {
			p.ClosePanels()
			escHandled = true
		}

		if escHandled {
			g.panelStack = append(g.panelStack[:idx], g.panelStack[idx+1:]...)
			g.updateLayout()
			break
		}
	}
}
```

However, this would require that every panel to these interfaces (`helpOverlay.Toggle` is used in the current code...)

d2game/d2player/move_gold_panel.go
```diff
@@ -0,0 +1,209 @@
...
+	incrose.SetPosition(moveGoldUpArrowX, moveGoldUpArrowY)
+	incrose.SetVisible(false)
+	incrose.OnActivated(func() { s.incrose() })
+	s.panelGroup.AddWidget(incrose)
+
+	decrose := s.uiManager.NewButton(d2ui.ButtonTypeDownArrow, d2resource.PaletteSky)
+	decrose.SetPosition(moveGoldDownArrowX, moveGoldDownArrowY)
+	decrose.SetVisible(false)
+	decrose.OnActivated(func() { s.decrose() })
+	s.panelGroup.AddWidget(decrose)
```
gravestench, line 123: `increase` and `decrease`

gravestench (approved):

d2game/d2player/game_controls.go
gravestench: I've opened an issue (#965) for refactoring this part of the code, I think that this PR is good to go.

essial:
Ill merge once conflict is resolved.

d2core/d2ui/textbox.go
gucio321: not, because, it is something like that: if we have e.g 25 gold, then if we type 25 it is ok, or?

gravestench: absolutely! i did not read the code enough, sorry.

PR #973, opened 2020-12-15
Escape menu mechanism

gucio321:
Hi there,
I've changed escape key effect in escape_menu to it looks like in original d2.
I've also changed save-file data (we don't need to store resistances in save file)
Also corrected minor logger bugs.


gucio321:
included in #975

PR #975, opened 2020-12-15
Add-skill buttons

gucio321:
Hi there,
In this PR I've added game-panel buttons, which gets activate, when we advance to a higher level. After we click on it, the appropiate panels are opened. Now we can also increase our stats with buttons next to it's fields.


gucio321:
So it's also a content from #973.

essial:
Can you pull logging logic out of this PR? There's another draft PR dealing with logging exclusively and I want to minimize scope.  Good work though!

PR #983, opened 2020-12-17
Quest Log - quest completion animation

gucio321:
Hi there,
Here is a list of major changes in this PR:
- when quest status = quest completing status quest completion animation is played
- fixed bug, when after switching to another tab, and getting back to previous we still was able to see highlighted socket
- fixed bug, when incorrect status value was passed to setQuestLabel, description was set to "qstsaXqX"
- added code descriptions
- removed still unused function questIDToCords


essial:
Are you on discord? If you are ping me and I'll tag you as a dev if you'd like.

PR #987, opened 2020-12-18
label-button widget

gucio321:
Hi there,
I've implemented label-button widget in d2ui (it works like d2gui.LabelButton). I think, it will be useful while creating e.g NPC's inventory.
Also implemented OnHover state in `d2ui.Label` and `d2ui.Sprite` (it didn't worked) .
I've added static checks for clickable widgets too.

### DIFF
--- a/d2core/d2ui/button.go
+++ b/d2core/d2ui/button.go
@@ -768,7 +768,8 @@ func getButtonLayouts() map[ButtonType]ButtonLayout {
 	}
 }
 
-var _ Widget = &Button{} // static check to ensure button implements widget
+var _ Widget = &Button{}          // static check to ensure button implements widget
+var _ ClickableWidget = &Button{} // static check to ensure button implemented clickable widget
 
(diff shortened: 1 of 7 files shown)


gravestench:
I think this should be split into multiple PR's

gravestench:
Yeah, you have 3 meaningful commits, but you're actually doing 4 things in this PR, and they're only related in that they are all in `d2ui`. Conceptually, the edits to these files were for different reasons, and those reasons are not reflected by the name you've given this PR.

Ultimately, it's up to @essial whether or not we allow merging PR's like this into master, but I'm against it.

I'm going to ask that you give this another attempt, but with 4 separate PR's for the following:
* Add static checks to all d2ui widgets
* Bugfix in d2ui for `Sprite.OnHover`
* Remove redundant `Label.GetSize` and `Sprite.GetSize`
* Add new `LabelButton` widget in d2ui

PR #988, opened 2020-12-18
Cleanup d2records logging

Intyre:
- [x] Using log level Debug instead of Info
- [x] Renamed some structs

### DIFF
--- a/d2core/d2records/set_loader.go
+++ b/d2core/d2records/set_loader.go
@@ -107,7 +107,7 @@ func setLoader(r *RecordManager, d *d2txt.DataDictionary) error {
 
 	r.Item.Sets = records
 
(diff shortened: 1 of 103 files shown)


d2core/d2records/set_loader.go
```diff
@@ -107,7 +107,7 @@ func setLoader(r *RecordManager, d *d2txt.DataDictionary) error {
 
 	r.Item.Sets = records
 
-	r.Logger.Infof("Loaded %d records records", len(records))
+	r.Debugf("Loaded %d Set records", len(records))
```
gravestench, line 110: good catch! i probably fucked this up during a refactor.

gravestench (approved):

PR #989, opened 2020-12-19
added label-button widget

gucio321:
Hi there,
in this PR I've added label-button widget to d2ui

### DIFF
--- a/d2core/d2ui/label_button.go
+++ b/d2core/d2ui/label_button.go
@@ -1,0 +1,121 @@
+package d2ui
+
+import (
+	"image/color"
+
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2interface"
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2util"
+)
+
+const (
+	white = 0xffffffff
+)
+
+// static checks to ensure LabelButton implemented Widget and ClickableWidget
+var _ Widget = &LabelButton{}
+var _ ClickableWidget = &LabelButton{}
+
+// LabelButton represents LabelButton
+type LabelButton struct {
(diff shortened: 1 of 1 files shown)


d2core/d2ui/label_button.go
```diff
@@ -0,0 +1,121 @@
...
+// LabelButton represents LabelButton
+type LabelButton struct {
+	*BaseWidget
+	stdColor   color.Color
+	hoverColor color.Color
+	onClick    func()
+	label      *Label
+
+	*d2util.Logger
+}
```
gravestench, line 27: Is it more coherent to store multiple `BaseWidget` references or just use the one inside of Label... ?

Currently, you're adding an anonymous field for `BaseWidget`, as well as adding a `Label` which also has an anonymous `BaseWidget` field.
 
Correct me if I'm wrong, but I think something like this should still work:
```golang
type LabelButton struct {
    *Label
	*d2util.Logger
	stdColor   color.Color
	hoverColor color.Color
	onClick    func()
}
```

then, your methods would end up looking like this:
```golang
func (b *LabelButton) SetText(text string) {
	b.Label.SetText(text)
	b.width, b.height = b.Label.GetSize() // uses b.Label.BaseWidget fields
}
```

gucio321: Yeah. You're right, it could be better, but sadly doesn't work. 
problem:
now I've separated label position and widget position.
I'm using HorizontalAlignCenter (has it's own XOffset)
So now, if I'll merge Label.BaseWIdget with BaseWidget real actionable label's region will start in 1/2 of label and ends 1/2 of label on it's right side.

gravestench (approved):

PR #991, opened 2020-12-19
fixed onHover bug in d2ui.Label

gucio321:
Hi there,
there was the bug in d2ui.Label. If we tried to use OnHoverStart and OnHoverEnd for Label, nothing happen


gravestench (approved):

gravestench:
Thank you!

Issue #965, opened 2020-12-07
Refactor the way panels are opened/closed


gravestench:
We should refactor the way we handle opened UI panels. The code for this is in `d2game/d2player/game_controls.go`. In particular, look towards `GameControls.onEscKey`.

I am thinking that we should implement a stack in `GameControls` for managing the panels.

It could be used like this:
```golang
type panel interface {
	IsOpen() bool
	Close()
}

type panelContainer interface {
	HasOpenPanels() bool
	ClosePanels()
}

func (g *GameControls) onEscKey() {
	// special case for opening the escape menu when panel stack is empty
	if len(g.panelStack) == 0 {
		g.escapeMenu.Open()

		g.panelStack = append(g.panelStack, g.escapeMenu)

		g.updateLayout()

		return
	}

	// generic case, pop the last thing off the, close it, and then update
	lastIdx := len(g.panelStack)-1
	last := g.panelStack[lastIdx]

	if p, isPanel := last.(panel); isPanel {
		p.Close()
	}

	if p, isPanelContainer := last.(panelContainer); isPanelContainer && {
		p.ClosePanels()
	}

	g.panelStack = append(g.panelStack[:lastIdx], g.panelStack[lastIdx+1:]...)
	g.updateLayout()
}
```

krishnaindani:
hi, Is this issue available? I would like to work on this @gravestench

gravestench:
Okay, assigned to you.

krishnaindani:
Thanks, I will get started on this.

gravestench:
Status?

gravestench:
I don't think @krishnaindani has even forked the od2 repo, they are no longer assigned.

krishnaindani:
Sorry, I got stucked with something else.

gucio321:
@gravestench I'm working on it now

PR #996, opened 2020-12-21
skill select menu dependencies

gucio321:
Hi there,
I'm adding several dependencies for skillselect panels.
In retail d2, if we open any of skill panels and then, we open any of game panels (e. g. inventory) the skillselect menu is closed.
And conversely, if we open inventory panel, and then skill select menu, inventory is closed.

### DIFF
--- a/d2game/d2player/game_controls.go
+++ b/d2game/d2player/game_controls.go
@@ -633,6 +633,8 @@ func (g *GameControls) OnMouseButtonDown(event d2interface.MouseEvent) bool {
 
 func (g *GameControls) toggleHeroStatsPanel() {
 	if !g.HelpOverlay.IsOpen() {
+		g.hud.skillSelectMenu.LeftPanel.Close()
+		g.hud.skillSelectMenu.RightPanel.Close()
 		g.questLog.Close()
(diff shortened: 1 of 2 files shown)


d2game/d2player/game_controls.go
```diff
@@ -645,9 +647,31 @@ func (g *GameControls) onCloseHeroStatsPanel() {
...
+func (g *GameControls) toggleRightSkillPanel() {
+	if !g.HelpOverlay.IsOpen() {
+		g.inventory.Close()
+		g.skilltree.Close()
+		g.questLog.Close()
+		g.heroStatsPanel.Close()
+		g.hud.skillSelectMenu.ToggleRightPanel()
+	}
+}
+
```
gravestench, line 669: For now this is fine, but I don't like how the logic is hard-coded in this file. The operation of showing/hiding panels should be _way_ more generic (#965).

When we add the charsi-imbue, or the Larzuk-socketing panels we also need to update all of the panel open/close logic...? Seems wrong, no?

gucio321: Yeah, I completely agree with you. I just want to point, what should be here included.

PR #1001, opened 2020-12-22
Game control refactor

gucio321:
Hi there,
here is refactor of game controls (#965)

### DIFF
--- a/d2game/d2player/game_controls.go
+++ b/d2game/d2player/game_controls.go
@@ -28,7 +28,6 @@ const (
 // Panel represents the panel at the bottom of the game screen
 type Panel interface {
 	IsOpen() bool
-	Toggle()
 	Open()
 	Close()
 }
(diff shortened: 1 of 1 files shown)


d2game/d2player/game_controls.go
```diff
@@ -28,7 +28,6 @@ const (
 // Panel represents the panel at the bottom of the game screen
 type Panel interface {
 	IsOpen() bool
-	Toggle()
```
gucio321, line 31: what do you think, if I'm removing it, should I remove Toggle() function from all panels' widgets?

d2game/d2player/game_controls.go
```diff
@@ -632,106 +588,111 @@ func (g *GameControls) OnMouseButtonDown(event d2interface.MouseEvent) bool {
...
-		g.hud.miniPanel.SetMovedRight(g.questLog.IsOpen())
-		g.updateLayout()
-	}
+	g.openLeftPanel(g.questLog)
 }
 
 func (g *GameControls) onCloseQuestLog() {
-	g.hud.miniPanel.SetMovedRight(g.questLog.IsOpen())
-	g.updateLayout()
 }
```
gucio321, line 668: should I leave it here?

ianling:
For what it's worth, I compiled the changes and tested and it looks great. Opening and closing one panel at a time, the viewport still changes properly. Opening and closing multiple panels works as expected. Pressing escape clears the screen of all panels, or opens the pause menu if no panels were open. Also tried rapidly opening and closing panels on both sides to try to break it and it seems very resilient.

d2game/d2player/game_controls.go
ianling: I tested quickly, and it looks like you would also need to add a check for if the onCloseCb is nil if you remove these empty functions. `Close()` doesn't check if the function pointer is nil before trying to run it.

I think it might save someone some work in the future to leave these empty functions here, in case they need to use them, but I also can't think of a situation where these would be needed, so maybe it's best for them just to go away.

d2game/d2player/game_controls.go
ianling: I vote yes

Issue #396, opened 2020-06-22
Support client disconnects in the networking layer


essial:
When a client disconnects, we need to detect this and either remove them from the game (in the case of a client connection), or disconnect all of the clients (in the case of the host leaving). For now it is acceptable to just close out the game with a message printed until we support leaving the game back to the main menu.

stephenhoran:
Could this be assigned to me? I have some of this sorted out, hopefully :).

essial:
Sure thing! Just message back if you give up on this so I can un-assign. Good luck!

gravestench:
@stephenhoran did you make any progress?

stephenhoran:
@gravestench the network stack does handle disconnects but looking at this now it do not remove the disconnected user sprite from the game.

ianling:
I can take this one

PR #1008, opened 2020-12-28
Networking bugfixes and cleanup

### LINKED ISSUE #877: Remove client player map entity on client disconnect
When a game client disconnects, the game server should remove the game client entity.

#396 is related

ianling:
Closes #877 and #396 

Changes:

- Make sure connections close properly, without weird error messages on either end.
- Remove player map entity when a player disconnects from a multiplayer game (#877)
- Close server properly when host disconnects, and handle ServerClose on remote clients (#396)
- Don't mix JSON Decoders and raw TCP writes. Use JSON Encoders
- Actually handle incoming packets from remote clients. They were basically being discarded before.
- General code cleanup for simplicity and consistency. Removes a bunch of duplicate code


gravestench (approved):
Well done!

PR #1014, opened 2020-12-30
Removed improper ebiten dependency in d2interface.

essial:


gravestench (approved):

gravestench:
For others looking at this PR, this is quick and dirty to remove ebiten as a dependency from `d2common/d2interface/`. Without this fix we were unable to re-use some code in hellspawner.

PR #1015, opened 2021-01-01
Render HUD before Panels 

### LINKED ISSUE #936: NPC hover-label is displayed above UI panels (eg. inventory)
**Describe the bug**
 If NPC walks under the inventory window, we still can see his label

**To Reproduce**
1. Go to inventory window
2. If your mouse will be over the NPC, which walks under the menu, you can see its label

**Additional context**
That's minor bug

ThomasChr:
(in this Case 'Panels' only does mean Inventory Panel). This is to avoid Entity Labels to be rendered above the Inventory Panel. Fixes #936


ThomasChr:
I tested and did not find any regressions like inventory panel beeing in front of the lower HUD or the mana globe.

gravestench (approved):

Issue #977, opened 2020-12-16
Fix all lint issues on ECS branch


gravestench:

ianling:
I am handling this

gravestench:
Resolved by #1017

PR #1017, opened 2021-01-03
WIP: Add checkboxes, de-lint ecs branch

### LINKED ISSUE #979: implement `UIWidgetSystem.Checkbox`
The method should have the signature:
```golang
func (t *UIWidgetSystem) Checkbox(checkedState bool) akara.EID
```

The sprite should be created using `UIWidgetSystem.spriteFactory.Sprite` 

the sprite path and palette should be hardcoded to: 
* d2resource.Checkbox
* d2resource.PaletteFechar

ianling:
Closes #979 and #977.

I would click on the individual commits to see changes from the checkbox issue specifically. The checkbox stuff was slightly altered in the de-lint commit, but stayed mostly the same.


d2core/d2checkbox/checkbox.go
```diff
@@ -0,0 +1,155 @@
...
+	if v.GetEnabled() {
+		v.Toggle()
+	}
+
+	if v.callback != nil {
+		return v.callback(thisComponent)
+	}
+
+	return false
+}
```
gravestench, line 107: Why does this method need an argument?
would it make more sense to pass `v` to `v.callback` as the reference to `thisComponent` ?
```golang
	if v.callback != nil {
		return v.callback(v)
	}
```

d2core/d2checkbox/checkbox.go
```diff
@@ -0,0 +1,155 @@
...
+	case v.GetEnabled():
+		// unchecked, enabled
+		_ = v.Sprite.SetCurrentFrame(0)
+	case v.GetPressed():
+		// checked, disabled
+		_ = v.Sprite.SetCurrentFrame(1)
+	default:
+		// unchecked, disabled
+		_ = v.Sprite.SetCurrentFrame(0)
+	}
```
gravestench, line 154: it looks like this switch can be boiled down a bit more, but maybe i'm not looking at it correctly... or maybe it just makes a lot more sense to keep all of the cases separate for readability.

d2core/d2systems/scene_ebiten_splash.go
```diff
@@ -105,7 +107,7 @@ func (s *EbitenSplashScene) createSplash() {
 	size := 10
 
 	totalW, totalH := len(flags[0])*size, len(flags)*size
-	ox, oy := (800-totalW)/2, (600-totalH)/2
+	ox, oy := (800-totalW)/2, (600-totalH)/2 //nolint:gomnd // halving things...
```
gravestench, line 110: i suppose i should have just done `>>1` here

d2core/d2checkbox/checkbox.go
ianling: In general, I think it would be useful to have access to the object itself in your callback. I think this is how a lot of callback  and event-based stuff in Javascript works.

I think it would be a lot cleaner looking to just pass `v` in to the callback, but that does not give developers as much control, since `v` is the 'inner' `d2checkbox.Checkbox` type and not a `d2component.Checkbox`. The latter (the way it is currently written) gives you the EID of the checkbox so there's more that you are able to do within the callback. I think that is preferable, but it's definitely kind of dirty looking.

Pending your thoughts, if we want to continue passing the outer component in, we could move `Activate` and some of these other functions to the outer `d2component.Checkbox` (and do that for other components as well, like buttons), and that would give us the ability to easily pass it into the callback without weird hacky workarounds.

d2core/d2checkbox/checkbox.go
ianling: This is where I was messing with the setting the sprite colors and effects, to properly render a "disabled" checkbox. I couldn't get anything working, so I ended up stripping it out, but my thought was that if the checkbox is disabled, the color should be changed to silver or something, like it is in retail D2

d2core/d2systems/scene_ebiten_splash.go
ianling: :shrug: either way, I think the way you did it helps keep it clearer for people unfamiliar with bitwise stuff

d2core/d2checkbox/checkbox.go
ianling: I personally prefer to keep the cases separate in case this eventually gets added in the future (and it's less work for me!)

d2core/d2checkbox/checkbox.go
gravestench: passing the EID around makes great sense, good point!

I don't really have any issues with the code as it is, i just wanted to discuss.

d2core/d2checkbox/checkbox.go
gravestench: that's a good reason to keep these cases separate, i don't have any issues with that.

d2core/d2systems/scene_ebiten_splash.go
gravestench: true. I've started doing `>>1` in other code to satisfy the lint gods without having to write lint directives.

gravestench (approved):

PR #1025, opened 2021-01-10
Fixed various bugs, crashes, and slowdowns.

essial:


d2common/d2loader/filesystem/loader_provider.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"github.com/OpenDiablo2/OpenDiablo2/d2common/d2loader/asset"
+)
+
+// OnAddSource is a shim method to allow loading of filesystem sources
+func OnAddSource(path string) (asset.Source, error) {
+	return &Source{
+		Root: path,
+	}, nil
+}
```
gravestench, line 12: Why do we need this?

d2core/d2asset/asset_manager.go
```diff
@@ -474,6 +477,99 @@ func (am *AssetManager) commandAssetClear([]string) error {
...
 
 	return nil
 }
+
+func (am *AssetManager) LoadDT1(dt1Path string) (*d2dt1.DT1, error) {
+	if dt1Value, found := am.dt1s.Retrieve(dt1Path); found {
+		return dt1Value.(*d2dt1.DT1), nil
+	}
+
+	fileData, err := am.LoadFile("/data/global/tiles/" + dt1Path)
```
gravestench, line 493: we may want to put these paths into `d2resource` at some point. I think there are a couple more used in the audio engine.

gravestench (approved):

PR #1031, opened 2021-01-16
Party Panel

gucio321:
Hi there,
That's initial of multi-player's party panel.


gucio321:
What contains this PR?:
- party panel frame, panel, close button e.t.c
- we can see here another players, when they're joining to the game
- labels are colored into appropriate colors (reflacts players relations)
- when our's level, or another player's level is less then 9, 'Go Hostile" button is disabled
- we can switch panel's switchers
- multiple tooltips on buttons, players' names e.t.c are there too

essial (approved):

PR #1033, opened 2021-01-23
Fix viewport and minipanel positions when closing panels via UI button

cardoso:
The viewport and minipanel had wrong positions when you close a panel via its UI button (the bug doesn't happen when closing via keyboard buttons)

To fix this:
- Moved the minipanel positioning logic to updateLayout & created miniPanel.ResetPosition function
- Called updateLayout in every panel's onClose callback

### DIFF
--- a/d2game/d2player/game_controls.go
+++ b/d2game/d2player/game_controls.go
@@ -621,6 +621,7 @@ func (g *GameControls) togglePartyPanel() {
 }
 
 func (g *GameControls) onCloseHeroStatsPanel() {
+	g.updateLayout()
 }
 
 func (g *GameControls) toggleLeftSkillPanel() {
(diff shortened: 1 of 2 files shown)


d2game/d2player/game_controls.go
```diff
@@ -621,6 +617,7 @@ func (g *GameControls) togglePartyPanel() {
 }
 
 func (g *GameControls) onCloseHeroStatsPanel() {
+	g.updateLayout()
```
gucio321, line 620: Good catch,
I had to miss it in my game controls refactor,
maybe, because I used to open panels with hotkeys :grinning:

gucio321 (approved):
Good catch,
I had to miss it in my game controls refactor,
maybe, because I used to open panels with hotkeys :grinning:

essial (approved):

PR #1036, opened 2021-01-27
circleci update

gucio321:


gucio321:
test filed, becaouse OpenDiablo2:master contains lint errors

PR #1045, opened 2021-02-05
Minor refactor of d2cof

gravestench:
* Changed `Load` to `Unmarshal`
* made `Marshal` and `Unmarshal` into methods of `COF`
* added `New` function which creates a new, empty COF instance
* added helper functions `Marshal` and `Unmarshal` (separate from the COF methods, these are package-level functions)
* Changed `StreamReader.ReadBytes` to account for edge case of reading 0
bytes (this was returning an error when it should not have)
* added really simple unit tests for COF

### DIFF
--- a/d2common/d2fileformats/d2cof/cof.go
+++ b/d2common/d2fileformats/d2cof/cof.go
@@ -8,9 +8,10 @@ import (
 )
 
 const (
-	unknownByteCount = 21
-	numHeaderBytes   = 4 + unknownByteCount
-	numLayerBytes    = 9
+	numUnknownHeaderBytes = 21
+	numUnknownBodyBytes   = 3
+	numHeaderBytes        = 4 + numUnknownHeaderBytes
+	numLayerBytes         = 9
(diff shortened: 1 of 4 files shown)


d2common/d2fileformats/d2cof/cof.go
```diff
@@ -33,12 +34,41 @@ const (
...
+		CompositeLayers:    make(map[d2enum.CompositeType]int),
+		AnimationFrames:    make([]d2enum.AnimationFrame, 0),
+		Priority:           make([][][]d2enum.CompositeType, 0),
+	}
+}
+
+// Marshal a COF to a new byte slice
+func Marshal(c *COF) []byte {
+	return c.Marshal()
+}
```
gucio321, line 56: Is it necessary?
In my opinion we'll easier to use
```go
cof.Marshal()
```
instead of
```golang
d2cof.Marshal(cof)
```

essial (approved):

PR #1049, opened 2021-02-07
d2dc6 refactor + unit test for it

gucio321:


gravestench (approved):
thanks for the unit tests!

PR #1057, opened 2021-02-15
hotfix: ds1 editor: some fields need to be exported

gucio321:
Hi there,
we need to be able to set this value in another packages (in HellSpawner)


gucio321:
:angry: this also adds this *** space at the end of apt-get's command in circleci

essial (approved):

PR #1059, opened 2021-02-16
hotfix: ds1: method setupLayerTypes is now exported + circleci build job bugfix

gucio321:


gravestench (approved):

gravestench:
I think it's important to note here the reason why you are exporting this method: **Hellspawner**

PR #1066, opened 2021-02-17
Refactoring d2ds1

gravestench:
This is a large refactor to the DS1 struct. Basically, the DS1 used to maintain layers per-tile, and this refactor makes the DS1 contain layers, and the layer contain tiles.

All of this was done because if we wanted to add layers to a tile, we would need a bunch of state management code to go ensure that all other tiles also have those layers (this would have been nasty and bad).

Also, this is pretty much all done for Hellspawner, since we're at a point where we want to make an editor for DS1's and we need to be able to add/edit/create/delete layers.

~~**We still need to finish the unit tests for `d2common/d2fileformats/d2ds1`**~~


essial (dismissed):

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -39,34 +41,362 @@ const (
...
+	substitutionType           int32               // SubstitutionType (layer type): 0 if no layer, else type 1 or type 2
+	numberOfWallLayers         int32               // WallNum number of wall & orientation layers used
+	numberOfFloorLayers        int32               // number of floor layers used
+	numberOfShadowLayers       int32               // ShadowNum number of shadow layer used
+	numberOfSubstitutionLayers int32               // SubstitutionNum number of substitution layer used
+	// substitutionGroupsNum      int32               // SubstitutionGroupsNum number of substitution groups, datas between objects & NPC paths
+
+	dirty            bool // when modifying tiles, need to perform upkeep on ds1 state
+	unknown1         []byte
+	layerStreamTypes []d2enum.LayerStreamType
```
gucio321, line 61: I put it here while writing DS1.Marshal, but it wasn't good idea. we can just recalculate layer stream types while encoding.

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -331,7 +661,7 @@ func (ds1 *DS1) loadSubstitutions(br *d2datautils.StreamReader) error {
 func (ds1 *DS1) setupStreamLayerTypes() []d2enum.LayerStreamType {
```
gucio321, line 661: probably could be exported, because we'd like to know about streamLayerTypes (in HellSpawner)

d2common/d2fileformats/d2ds1/ds1.go
gucio321: solved by 5e62b12

d2common/d2fileformats/d2ds1/ds1_layers.go
```diff
@@ -0,0 +1,432 @@
...
+	substitutionLayerGroup
+)
+
+type layerGroup []*layer
+
+type ds1Layers struct {
+	width, height int
+	Floors        layerGroup
+	Walls         layerGroup
+	Orientations  layerGroup
```
gucio321, line 27: why did you splited it? IMO Wall and orientation should be one

et
```diff
@@ -0,0 +1,73 @@
...
+[1m+++ b/d2common/d2fileformats/d2ds1/ds1_test.go[m
+[36m@@ -575,6 +575,24 @@[m [mfunc TestDS1_SetNumberOfFloors(t *testing.T) {[m
+ 	if err := testIfRestorable(ds1, test); err != nil {[m
+ 		t.Errorf("unable to restore: %v", err)[m
+ 	}[m
+[32m+[m
+[32m+[m	[32mnewNumber = 10[m
+[32m+[m
+[32m+[m	[32mds1.SetNumberOfFloorLayers(newNumber)[m
+[32m+[m
```
gravestench, line 13: ew gross what is this ;_:

d2common/d2fileformats/d2ds1/ds1_layers.go
gravestench: yeah, you're right

d2common/d2fileformats/d2ds1/ds1_layers.go
```diff
@@ -0,0 +1,432 @@
...
+		group = &l.Walls
+	case orientationLayerGroup:
+		group = &l.Orientations
+	case shadowLayerGroup:
+		group = &l.Shadows
+	case substitutionLayerGroup:
+		group = &l.Substitutions
+	default:
+		return nil
+	}
```
gucio321, line 214: maybe make this `switch-case` a method of ds1Layers.
like, 
```golang
func (l *ds1Layers) getLayerGroup(t layerGroupType) (group *layerGroup) {
switch t {
	case floorLayerGroup:
		group = &l.Floors
	case wallLayerGroup:
		group = &l.Walls
	case orientationLayerGroup:
		group = &l.Orientations
	case shadowLayerGroup:
		group = &l.Shadows
	case substitutionLayerGroup:
		group = &l.Substitutions
	default:
		return nil
	}
}

return group
```

d2common/d2fileformats/d2ds1/ds1_layers.go
```diff
@@ -0,0 +1,432 @@
...
+		idx = len(group) - 1
+	}
+
+	// example:
+	// suppose
+	//		idx=1
+	//		newLayer=c
+	// 		existing layerGroup is [a, b]
+	group = append(group, group[idx:]...) // [a, b] becomes [a, b, b]
+	group[idx] = newLayer                 // [a, b, b] becomes [a, c, b]
```
gucio321, line 296: what happens if:
idx=1
group = [a, b, c]
newLayer = d ?
then in append we have:
[a, b, c] + [b, c] so [a, b, c, b, c]
and after changing index:
[a, b, d, b, c]

but result of this operation should be:
[a, b, d, c]

so IMO this code will not work properly

gucio321: let's do just:
```golang
groups = append(groups[:idx], newLayer, groups[idx:]...)
```

d2common/d2fileformats/d2ds1/layer.go
```diff
@@ -0,0 +1,143 @@
...
+
+	Height() int
+	SetHeight(h int) grid
+
+	Size() (w, h int)
+	SetSize(w, h int) grid
+
+	Tile(x, y int) *Tile
+	SetTile(x, y int, t *Tile)
+}
```
gucio321, line 37: is this interface even necessary? you use it only in SetWIdth/Height/Size methods, and you could return simply `*layer` there

d2common/d2fileformats/d2ds1/ds1_layers.go
gucio321: fixed by 41c1d8e

d2common/d2fileformats/d2ds1/ds1_layers.go
gucio321: solved in 41c1d8e

et
gucio321: fixed by 41c1d8e

d2common/d2fileformats/d2ds1/ds1_layers.go
gucio321: fixed in 41c1d8e

d2common/d2fileformats/d2ds1/layer.go
gravestench: It isn't needed, I think this is just here because i used goland to write the boilerplate code from this interface, will remove.

d2common/d2fileformats/d2ds1/ds1.go
gravestench: `ds1.Floors`, `ds1.Walls`, `ds1.Shadows`, and `ds1.Substitutions` are now public, we should be able to get this information without a method like this being exported.

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -39,232 +27,252 @@ const (
...
-	unknown1                   []byte
-	LayerStreamTypes           []d2enum.LayerStreamType
-	unknown2                   uint32
-	NpcIndexes                 []int
+	*ds1Layers
+	Files              []string            // FilePtr table of file string pointers
+	Objects            []Object            // Objects
+	substitutionGroups []SubstitutionGroup // Substitution groups for the DS1
+
+	version          ds1version
```
gucio321, line 35: we need to be able to set version

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -39,232 +27,252 @@ const (
...
+	stream := d2datautils.CreateStreamReader(fileData)
 
-	if ds1.Version >= v9 && ds1.Version <= v13 {
-		// Skipping two dwords because they are "meaningless"?
-		ds1.unknown1, err = br.ReadBytes(unknown1BytesCount)
-		if err != nil {
-			return nil, err
-		}
+	if err := ds1.loadHeader(stream); err != nil {
+		return nil, fmt.Errorf("loading header: %v", err)
```
gucio321, line 60: could you use error-wrapping expression (`%w`)

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -273,94 +281,108 @@ func (ds1 *DS1) loadObjects(br *d2datautils.StreamReader) error {
...
+	var layerStream []layerStreamType
 
-	if ds1.Version < v4 {
-		layerStream = []d2enum.LayerStreamType{
-			d2enum.LayerStreamWall1,
-			d2enum.LayerStreamFloor1,
-			d2enum.LayerStreamOrientation1,
-			d2enum.LayerStreamSubstitute,
-			d2enum.LayerStreamShadow,
+	if ds1.version < v4 {
```
gucio321, line 343: let's do something like `version.standardLayers()`

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -434,20 +453,20 @@ func (ds1 *DS1) loadNpcPaths(br *d2datautils.StreamReader, objIdx, numPaths int)
...
 		py, err := br.ReadInt32() //nolint:govet // i want to re-use the err variable...
 		if err != nil {
-			return err
+			return fmt.Errorf("reading Y point for path %d: %v", pathIdx, err)
 		}
 
 		newPath.Position = d2vector.NewPosition(float64(px), float64(py))
 
-		if ds1.Version >= v15 {
+		if ds1.version >= v15 {
```
gucio321, line 466: ds1Version.dederminesNPCAction()

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -456,58 +475,62 @@ func (ds1 *DS1) loadNpcPaths(br *d2datautils.StreamReader, objIdx, numPaths int)
...
 	}
 
-	return err
+	return nil
+}
+
+// SetSize sets the size of all layers in the DS1
+func (ds1 *DS1) SetSize(w, h int) {
+	ds1.ds1Layers.SetSize(w, h)
 }
```
gucio321, line 534: is it necessary?

d2common/d2fileformats/d2ds1/layer.go
```diff
@@ -0,0 +1,129 @@
...
+	layerStreamShadow1
+	layerStreamSubstitute1
+)
+
+type tileRow []Tile     // index is x coordinate
+type tileGrid []tileRow // index is y coordinate
+
+type layer struct {
+	tiles tileGrid
+}
```
gucio321, line 27: lol, why did you make it as a two types? isn one not enough?

d2common/d2fileformats/d2ds1/ds1.go
gravestench: yes

d2common/d2fileformats/d2ds1/layer.go
gravestench: eh, for clarity; "a grid is a slice of rows"

d2core/d2config/defaults.go
```diff
@@ -20,7 +20,7 @@ func DefaultConfig() *Configuration {
 		VsyncEnabled:    true,
 		SfxVolume:       defaultSfxVolume,
 		BgmVolume:       defaultBgmVolume,
-		MpqPath:         "C:/Program Files (x86)/Diablo II",
+		MpqPath:         "C:/Program files (x86)/Diablo II",
```
gucio321, line 23: why did you changed it?

gravestench: probably on accident

main.go
```diff
@@ -8,11 +8,11 @@ import (
 
 // GitBranch is set by the CI build process to the name of the branch
 //nolint:gochecknoglobals // This is filled in by the build system
-var GitBranch string = "local"
+var GitBranch = "local"
 
 // GitCommit is set by the CI build process to the commit hash
 //nolint:gochecknoglobals // This is filled in by the build system
-var GitCommit string = "build"
+var GitCommit = "build"
```
gucio321, line 15: also here, why did you changed it? and if so, lets make just `GitBranch := "local"`

gravestench: I think just because the `string` type declaration was unnecessary.

you can't use `:=` where these variables are being declared.

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -39,232 +27,252 @@ const (
...
-	unknown2                   uint32
-	NpcIndexes                 []int
+	*ds1Layers
+	Files              []string            // FilePtr table of file string pointers
+	Objects            []Object            // Objects
+	substitutionGroups []SubstitutionGroup // Substitution groups for the DS1
+
+	version          ds1version
+	Act              int32 // Act, from 1 to 5. This tells which Act table to use for the Objects list
+	substitutionType int32 // SubstitutionType (layer type): 0 if no layer, else type 1 or type 2
```
gucio321, line 37: also, what with this? Could you add setter/getter or just export

gravestench: I'm not sure what it is used for, I'll export anyways.

d2common/d2fileformats/d2ds1/ds1.go
```diff
@@ -39,232 +27,252 @@ const (
...
+	unknown1         []byte
```
gucio321, line 38: also, could you do something with this?
I mean, that file created with &DS1{}.Marshal() will be incorrect
so maybe replace with `[unknown1BytesCount]byte` or something like that

gucio321: hmm. thats 8 bytes, so int64.

main.go
gucio321: > you can't use := where these variables are being declared.

oh, ok, I forgot, sorry

d2common/d2fileformats/d2ds1/ds1_layers_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+func Test_ds1Layers_DeleteWall(t *testing.T) {}
+
+func Test_ds1Layers_GetFloor(t *testing.T) {}
+
+func Test_ds1Layers_GetShadow(t *testing.T) {}
+
+func Test_ds1Layers_GetSubstitution(t *testing.T) {}
+
+func Test_ds1Layers_GetWall(t *testing.T) {}
```
gucio321, line 21: this tests should be completed

gucio321 (changes requested):
I think, that the tests should be written before merging

d2common/d2fileformats/d2ds1/ds1_layers_test.go
gravestench: setting back to draft

gucio321 (approved):
Great work! Thanks for doing it!

gucio321 (approved):

PR #1070, opened 2021-02-19
hotfix: font format:

gucio321:
- removed magic numbers
- corrected unknown1 bytes count in d2fontglyph.Create


gucio321:
test passed:
image

gravestench (approved):

Issue #909, opened 2020-11-08
File Encoders for all file formats in `d2common/d2fileformats/


gravestench:
We currently only have file decoders in `d2fileformats`. Most of the `Load` functions take a byte slice as an argument, and parse the bytes into a struct for that file type. We need to be able to convert a struct instance back to a byte slice.

rjnienaber:
I did a cursory check and it seems like `d2dcc` and `d2mpq` are the only file formats that still require a `.Marshal` method. Is this required for those formats? If so, I'd like to take a shot at the `.dcc` one but I don't have an example file that I could commit to the repo.

PR #1075, opened 2021-02-25
general refactor of text dictionary encoder

### LINKED ISSUE #1054: editing string tables
Already, we have data encoder in d2common/d2fileformats/d2tbl, but we are still unable to edit it. We need to write `AddEntry()` and `DeleteEntry()` methods for TextDIctionary.

reffers to #909 , #1040 and https://github.com/OpenDiablo2/HellSpawner/issues/86

gucio321:
Hi there,
written by me text dictionary encoder was extremely ineffectual, because we was unable to add and remove entries from table.
In this PR I've completely removed my previous encoder and written `d2tbl.Marshal` method again.
also, this fixes #1054, because now we can simply
```golang
if _, ok := dict[key]; !ok {
     dict[key] = newValue
} 
```
this is also related with #909 and https://github.com/OpenDiablo2/HellSpawner/issues/86
the only weak point of this solution is the inability to use newly coded dictionaries in original d2

In this PR I've also added more descriptive error mesages to LoadTextDictionary

it adds unit test for text dictionary too


gucio321:
image

gravestench (approved):

gravestench:
> the only weak point of this solution is the inability to use newly coded dictionaries in original d2

Why?

gucio321:
there are several fields, where I just push `0`, because our decoder doesn't use thes bytes. however it **might** be important in retail D2.

gucio321:
other words:
current encoder:
\+ encoded data **probably** could be used in original D2
\- realy hard to edit
\- many redundant fields in dictionary's structure

this PR:
\+ realy independent encoder (in fact could be used on any `map[string]string`)
\+ extremely easy to edit (just add key-value to map)
\- **probably** cannot use in retail D2

PR #1078, opened 2021-02-26
animation data engine + encoder

gucio321:
Hi there,
We had two animation data managers (one in `d2data`, second in `d2fileformats`), but one (in `d2fileformats`) was completely unused.
in this PR I've changed code, to use this second anim data manager (IMO is more advanced) and written encoder for it.
related with #909


gucio321:
image

gravestench (approved):

PR #1080, opened 2021-02-28
bugfix string table encoding

gucio321:
I've used map ranges twice, and each loop ranged the map in another order, so text dictionary encoder sometimes didn't work


gucio321:
image

gucio321:
image

gravestench (approved):

PR #1082, opened 2021-03-01
bugfix: d2util.Logger.Fatal()

gucio321:
Hi there,
d2util.Logger.Fatal() didn't terminate program, when d2util.Logger.level < LogLevelFatal


gravestench (approved):
nice

PR #1086, opened 2021-03-04
animation data: methods for edition animation data records count and animation entries

gucio321:
Hi there,
this PR contains:
- Push Record (adds a new record to specific entry
 - Delete Record (removes specified record from specified entry)
 - Add Entry (adds a new entry with specific name)
 - Delete Entry (removes specified entry)
- unit tests for these methods

necessary for OpenDiablo2/HellSpawner#78
and to complete OpenDiablo2/HellSpawner#219


gucio321:
image

essial (approved):

Issue #907, opened 2020-11-08
Unit Test: `d2common/d2cache`


gravestench:
we need unit tests for each method of `d2cache.Cache`

PR #1093, opened 2021-03-21
Anvil of fury/cache tests

anvil-of-fury:
resolves: https://github.com/OpenDiablo2/OpenDiablo2/issues/907

Only question I have is do we want to allow the cache to insert a node with `weight` > `budget`? Please see `TODO` comment for this test case and let me know if this is expected behavior.

### DIFF
--- a/d2common/d2cache/cache_test.go
+++ b/d2common/d2cache/cache_test.go
@@ -7,15 +7,16 @@ import (
 func TestCacheInsert(t *testing.T) {
 	cache := CreateCache(1)
 	insertError := cache.Insert("A", "", 1)
+
 	if insertError != nil {
 		t.Fatalf("Cache insert resulted in unexpected error: %s", insertError)
 	}
 }
 
-// TODO: do we want cache insert to fail if we add a node with weight > budget?
 func TestCacheInsertWithinBudget(t *testing.T) {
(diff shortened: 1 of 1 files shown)


anvil-of-fury:
I had to remove a TODO comment due to linting error, but here is the test in my above question: https://github.com/OpenDiablo2/OpenDiablo2/pull/1093/files#diff-689b9048057e7ac27ccbe8e5ef6ef87bab49be8baa8e33102af9f14712ead4f2R16-R23

d2common/d2cache/cache_test.go
```diff
@@ -0,0 +1,93 @@
...
+func TestCacheInsert(t *testing.T) {
+	cache := CreateCache(1)
+	insertError := cache.Insert("A", "", 1)
+
+	if insertError != nil {
+		t.Fatalf("Cache insert resulted in unexpected error: %s", insertError)
+	}
+}
+
+func TestCacheInsertWithinBudget(t *testing.T) {
```
anvil-of-fury, line 16: do we want to allow nodes to be inserted if their `weight` > `budget`?

gucio321 (approved):
good work.

essial (approved):

PR #1098, opened 2021-03-24
Use filepath instead of path for Windows support

willroberts:
The current build fails to create the `OpenDiablo2` directory on Windows without this change.

In contrast to `path`, `path/filepath` is OS-aware and handles Windows path separators correctly.

### DIFF
--- a/d2common/d2fileformats/d2mpq/mpq.go
+++ b/d2common/d2fileformats/d2mpq/mpq.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -201,5 +200,5 @@ func openIgnoreCase(mpqPath string) (*os.File, error) {
 		}
 	}
 
-	return os.Open(path.Join(mpqDir, mpqName)) //nolint:gosec // Will fix later
+	return os.Open(filepath.Join(mpqDir, mpqName)) //nolint:gosec // Will fix later
 }
--- a/d2core/d2hero/hero_state_factory.go
+++ b/d2core/d2hero/hero_state_factory.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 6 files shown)


drpaneas (dismissed):
/lgtm

d2core/d2hero/hero_state_factory.go
```diff
@@ -224,15 +223,15 @@ func (f *HeroStateFactory) getGameBaseSavePath() (string, error) {
 		return "", err
 	}
 
-	return path.Join(configDir, "OpenDiablo2/Saves"), nil
+	return filepath.Join(configDir, "OpenDiablo2/Saves"), nil
```
drpaneas, line 226:
```suggestion
	return filepath.Join(configDir, "OpenDiablo2", "Saves"), nil
```

drpaneas (changes requested):
See my comment, you need to change the syntax a bit ;)

d2common/d2fileformats/d2mpq/mpq.go
```diff
@@ -201,5 +200,5 @@ func openIgnoreCase(mpqPath string) (*os.File, error) {
 		}
 	}
 
-	return os.Open(path.Join(mpqDir, mpqName)) //nolint:gosec // Will fix later
+	return os.Open(filepath.Join(mpqDir, mpqName)) //nolint:gosec // Will fix later
```
gucio321, line 203: could you also fix this nolint directive: https://stackoverflow.com/questions/52320708/how-to-handle-gosec-linter-warning-potential-file-inclusion-via-variable

gucio321 (approved):

d2game/d2gamescreen/credits.go
```diff
@@ -72,7 +72,7 @@ type Credits struct {
 
 // LoadContributors loads the contributors data from file
 func (v *Credits) LoadContributors() []string {
-	file, err := os.Open(path.Join("./", "CONTRIBUTORS"))
+	file, err := os.Open(filepath.Join(".", "CONTRIBUTORS"))
```
essial, line 75: Lol good catch

essial (approved):

PR #1100, opened 2021-03-24
Update golangci

drpaneas:
* Replace 'maligned' with 'fieldalignment' [1]
* Remove 'interfacer' [2]
* Enable linting against go test files as well
* Disable 'funlen' linter for go table-driven tests
* Disable fieldalignment

[1] https://github.com/golangci/golangci-lint/pull/1765
[2] https://github.com/golangci/golangci-lint/issues/541


.golangci.yml
```diff
@@ -23,11 +23,12 @@ linters-settings:
...
     min-confidence: 0.8
   govet:
+    enable:
+      - fieldalignment
     check-shadowing: true
+
   lll:
     line-length: 140
-  maligned:
-    suggest-new: true
```
gucio321, line 30: Is there something like `suggest-new: true` for govet? IMO it is hard to fix this lint errors without linter's suggestion.

drpaneas: Sadly there isn't any and fixing all of those stuff is not easy. Not to mention the struct readability goes away with this. I checked several projects and for now -- until the maintainers provide a better way to handle these pointer byte errors -- they have it disabled.

gucio321 (approved):

gucio321 (changes requested):
there is a new golangci-lint version and we have lint errors. (re-run workflow to see them)

essial (approved):

.golangci.yml
gucio321: yo, just found an easy way to fix `fieldalignment` lint errors
sadly it isn't available in `golangci-lint` yet
```sh
go get golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment
fieldalignment -fix ./...
```

PR #1111, opened 2021-03-29
d2compression/huffman.go - add break to fix false-positive of staticcheck lint error

gravestench:


gravestench:
i was getting lint errors from an old `golangci-lint` version, closing.

PR #1118, opened 2021-04-07
d2ds1: fixed bug, when the file was encoded incorrectly

gucio321:
Hi there,
when version was greater than 16, the file was encoded incorrectly


gravestench (approved):
oof

Issue #1119, opened 2021-04-17
Segmentation fault on Arch Linux


ShadowWarfare:
I've been trying to run OpenDiablo2 on Arch Linux which I got from AUR via the opendiablo2-git but after transferring all of the game files to the system and then running OpenDiablo2 from command line it gives the following error:
```

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x5572d832962b]

goroutine 1 [running, locked to thread]:
github.com/OpenDiablo2/OpenDiablo2/d2core/d2records.updateEquivalencies(0xc000e60ae0, 0xc000e68e40, 0xc0001d8d80, 0x0, 0x0, 0x0, 0x0)
	github.com/OpenDiablo2/OpenDiablo2/d2core/d2records/item_types_loader.go:128 +0x2b
github.com/OpenDiablo2/OpenDiablo2/d2core/d2records.LoadItemEquivalencies(0xc000e606c0, 0xc000e60ae0, 0xc00002f20f)
	github.com/OpenDiablo2/OpenDiablo2/d2core/d2records/item_types_loader.go:114 +0x445
github.com/OpenDiablo2/OpenDiablo2/d2core/d2records.itemTypesLoader(0xc000088c00, 0xc0004b2400, 0x5572d8652727, 0x20)
	github.com/OpenDiablo2/OpenDiablo2/d2core/d2records/item_types_loader.go:69 +0x1345
github.com/OpenDiablo2/OpenDiablo2/d2core/d2records.(*RecordManager).Load(0xc000088c00, 0x5572d8652727, 0x20, 0xc0004b2400, 0x0, 0x0)
	github.com/OpenDiablo2/OpenDiablo2/d2core/d2records/record_manager.go:300 +0xd6
github.com/OpenDiablo2/OpenDiablo2/d2core/d2asset.(*AssetManager).LoadRecords(0xc000302410, 0x5572d8652727, 0x20, 0x0, 0x0)
	github.com/OpenDiablo2/OpenDiablo2/d2core/d2asset/asset_manager.go:366 +0x85
github.com/OpenDiablo2/OpenDiablo2/d2app.(*App).initDataDictionaries(0xc0000c1540, 0xc000315c00, 0x0)
	github.com/OpenDiablo2/OpenDiablo2/d2app/initialization.go:127 +0xaf
[...]

PR #1122, opened 2021-04-30
fix panic while loading wrong version of game.

gucio321:
Hi there!
this PR should explain problems like in #1119

### DIFF
--- a/d2core/d2records/item_types_loader.go
+++ b/d2core/d2records/item_types_loader.go
@@ -1,6 +1,8 @@
 package d2records
 
 import (
+	"log"
+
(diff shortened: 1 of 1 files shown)


d2core/d2records/item_types_loader.go
```diff
@@ -125,6 +127,10 @@ func updateEquivalencies(
 	itemType *ItemTypeRecord,
 	checked []string,
 ) {
+	if itemType == nil {
+		log.Fatal("invalid data file. Please ensure, yo're using the newest patch_d2.mpq file!")
```
essial, line 131: you're

gucio321: 146f520

PR #1123, opened 2021-05-11
bitmuncher: add missing methods from stream reader

gucio321:


gravestench (dismissed):

gucio321:
I think, that I'll add something more here

d2common/d2datautils/bitmuncher.go
```diff
@@ -60,12 +63,12 @@ func (v *BitMuncher) SetBitsRead(n int) {
...
 // GetBit reads a bit and returns it as uint32
-func (v *BitMuncher) GetBit() uint32 {
-	result := uint32(v.data[v.offset/byteLen]>>uint(v.offset%byteLen)) & oneBit
+func (v *BitMuncher) GetBit() bool {
+	result := v.data[v.offset/byteLen] >> uint(v.offset%byteLen) & oneBit
 	v.offset++
 	v.bitsRead++
 
-	return result
+	return result == 1
```
gucio321: what do you think about it? result will be stored in a single-byte variable (instead of 4-byte uint32)

PR #1125, opened 2021-05-14
difficultyType: add fmt.Stringer implementation

gucio321:


d2common/d2enum/player_animation_mode_string.go
```diff
@@ -28,11 +28,12 @@ func _() {
 	_ = x[PlayerAnimationModeDead-17]
 	_ = x[PlayerAnimationModeSequence-18]
 	_ = x[PlayerAnimationModeKnockBack-19]
+	_ = x[PlayerAnimationModeNone-20]
 }
 
-const _PlayerAnimationMode_name = "DTNUWLRNGHTNTWA1A2BLSCTHKKS1S2S3S4DDGHGH"
+const _PlayerAnimationMode_name = "DTNUWLRNGHTNTWA1A2BLSCTHKKS1S2S3S4DDGHGH\"\" - aura skills, e.g. Paladin's Concentration Aura"
```
gravestench, line 34: is this intentional?

gucio321: hmm, yes and no, I re-generated enums. Should I revert this change?

gravestench: I'm assuming that `"" - aura skills, e.g. Paladin's Concentration Aura` is not an animation mode string, and it should not be in this file.

gucio321: so, the comment on PlayerAnimationModeNone is wrong:
`// "" - aura skills, e.g. Paladin's Concentration Aura`
should be over the constant, not in the same line.
I'm not sure, if it belongs to this PR, but I fixed it: eb6e7a0

PR #1135, opened 2021-06-13
Fieldalignment lintfix

gucio321:
Yo,
I found an easy way to fix `fieldalignment` lint errors an wanted to check what'll happend if I'll fix them all :grinning:.
I don't see any great performency improvements, but I suppose that the game should take a bit less memory now


gravestench:
You also changed all instances of `Id` to `ID`. Why isn't that in a separate PR?

gucio321:
6fb564a was because of linter alerts I suppose
I don't know why this alerts wasn't reported earlier.

